PR #12, opened 2018-11-19
Updating Home page and adding modal content

ryelo:
* Updates information inside the data table on home
* adds progress bar [wip]. Waiting until PF fixes layout issues. Then removing some classes
* started implementation on the modal. No data is being stored with it yet, just getting the flow working.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,5 @@
 {
-    presets: ["env", "react", "es2017", "es2015"],
+    "presets": ["env", "react", "es2017", "es2016", "es2015"],
     "plugins": [
         ["transform-runtime", {
             "polyfill": false,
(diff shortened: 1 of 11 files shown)


ryelo:
waiting on https://github.com/RedHatInsights/insights-frontend-components/pull/152 to merge

jharting:
@ryelo FYI I switched API representations to use snake case and pushed a commit here to align.

Also, it is a good coding practice to only use `let` when you know you will write to that variable later. Use `const` everywhere else - that way the code is easier to reason about. I pushed a commit that enables eslint's `prefer-const` rule and fixed the occurrences.

ryelo:
@jharting Awesome! Thanks! There should be one more commit here whenever the Wizard component merges that should add it. Then it should be ready to merge.

jharting:
@ryelo looks good. The only glitch I found is that I cannot type anything into the plan name input field. It looks editable but it does not work.

PR #14, opened 2018-11-29
Fixes typing inside of Create Modal naming input

ryelo:
#13

### DIFF
--- a/src/components/CreatePlanModal/ModalSteps/PlanName.js
+++ b/src/components/CreatePlanModal/ModalSteps/PlanName.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 
 import {
     Form,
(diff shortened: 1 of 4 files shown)


src/components/CreatePlanModal/ModalSteps/PlanName.js
```diff
@@ -1,33 +1,51 @@
...
-                >
-                    <TextInput
+class PlanName extends Component {
+
+    constructor () {
+        super();
+        this.state = {
+            value: ''
+        };
+        this.handleTextInputChange = this.handleTextInputChange.bind(this);
```
jharting: @ryelo wait, isn't the only reason why you are using class properties below to avoid having to do this binding?

ryelo: Yep! Totally missed that.

PR #18, opened 2018-12-05
remediations/id page

ryelo:
I know some of the system information won't be there for the first draft, but I have it in here just for filler. We can remove whenever necessary. 

TODO:
- [x] Breadcrumbs
- [x] Treeview Table

Blocked by: Patternfly table getting merged. Looks like it is really close. Once this is merged, I will swap out the tables which should also fix the current table bug.

We can add in the API whenever new things get added. Most things still waiting for the API are labeled "fix"


ryelo:
@jharting I'm going to be out Dec 6 + Dec 7. Feel free to take a look before then or make edits as you see fit.

src/routes/ListPlans.js
```diff
@@ -0,0 +1,142 @@
...
+    Card, CardHeader, CardBody,
+    Progress, ProgressMeasureLocation,
+    Stack, StackItem,
+    Switch,
+    Level, LevelItem
+} from '@patternfly/react-core';
+
+import './ListPlans.scss';
+
+class ListPlans extends Component {
```
jharting: This is more "RemediationDetails" page rather than a list, right?

jharting:
@ryelo looks good so far. The only nitpick is that word "plan" should not longer be used in any situation

jharting:
force-pushing to resolve merge conflicts

jharting:
@ryelo FYI I pushed a few enhancements

src/routes/ListPlans.js
ryelo: Correct. We can update the name.

ryelo:
@jharting I think this should be a food first go run with the page. 

Patternfly table needs to be added when it is merged and when new API is added, that will be added too.

PR #28, opened 2019-01-14
Updating red for 'reboot required' icon + text

ryelo:
For some reason, I had put in the default `red` instead of PF's red. Updated to reflect the true red color.


jharting:
@ryelo we have a similar thing here: https://github.com/RedHatInsights/insights-frontend-components/blob/master/src/SmartComponents/Remediations/steps/IssueResolutionStep.scss
I suggest that we create a reboot icon component (with label) in components and then reuse it in both places

ryelo:
@jharting Sounds good! I'll make it.

ryelo:
https://github.com/RedHatInsights/insights-frontend-components/pull/196

ryelo:
@jharting going to close this PR to better reflect the scope.

PR #31, opened 2019-01-16
Making hot loaded wizard buttons only on dev environments

ryelo:

PR #35, opened 2019-01-21
only show test buttons if remediations:debug localStorage entry is set

jharting:


jharting:
@ryelo Superseding https://github.com/RedHatInsights/insights-remediations-frontend/pull/31 with this as the test buttons are sometimes helpful even in prod envs. for testing.

PR #56, opened 2019-01-29
introduce SelectableTable and refactor existing Table sites

jharting:


src/containers/SelectableTable.js
```diff
@@ -0,0 +1,67 @@
...
+        if (!has(row, 'id')) {
+            return state;
+        }
+
+        if (!isSelected) {
+            return pickBy(state, (value, key) => key !== row.id);
+        }
+
+        return {
+            ...state,
```
karelhala: Do not spread state here, but `state.selected` to return already selected items.
```JS
return {
    ...state.selected,
    [row.id]: true
}
```

jharting:
@karelhala thanks, all should be fixed now

PR #70, opened 2019-02-08
Fixes/resolution modal

kruai:
Changes a few things in accordance with RHIPLAT2-400:

- Change some titles to title case
- Override "Confirm" button in modal to say "Save" instead
- Add loading skeleton while resolutions load in

### DIFF
--- a/src/components/ResolutionModal/ResolutionStep.js
+++ b/src/components/ResolutionModal/ResolutionStep.js
@@ -9,7 +9,8 @@ import {
 } from '@patternfly/react-core';
 
 import {
-    Reboot
+    Reboot,
+    Skeleton
 } from '@red-hat-insights/insights-frontend-components';
 
 import { getResolutions } from '../../api';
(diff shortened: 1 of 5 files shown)


jharting:
Waiting for https://github.com/RedHatInsights/insights-frontend-components/pull/292

src/components/ResolutionModal/ResolutionStep.js
```diff
@@ -45,45 +46,59 @@ class ResolutionStep extends Component {
...
+                    </StackItem>
+                    <StackItem>
+                        <Skeleton/>
+                    </StackItem>
+                    <StackItem>
+                        <Skeleton/>
+                    </StackItem>
+                </Stack>
+            );
+        } else {
```
jharting: <nitpick>Since the branch above uses return it is not necessary to use else here. Using else leads to more nesting which can be harder to read. https://eslint.org/docs/rules/no-else-return

kruai: OK! Cleaning that up now

kruai:
RedHatInsights/insights-frontend-components#292 has been merged; once this PR is reviewed I'll merge this in as well.

PR #76, opened 2019-02-12
remediation status indication

jharting:
* compliance not covered yet
* supposed to work in insightsbeta only for now


src/components/ActionsResolvedCard.js
```diff
@@ -0,0 +1,48 @@
...
+        value={ resolved }
+        max={ total }
+        label={ `${resolved} of ${total}` }
+        measureLocation={ ProgressMeasureLocation.outside } />;
+}
+
+const ActionsResolvedCard = ({ status }) => (
+    <Card className='ins-c-card__actions-resolved'>
+        <CardHeader>
+            <Level>
```
ryelo: Does this need to be a level? if we're following PF, we should just use "Actions Resolved" as the `title` prop.

jharting: @ryelo I am not following. I just split out this https://github.com/RedHatInsights/insights-remediations-frontend/blob/master/src/routes/RemediationDetails.js#L96 into a standalone component.

ryelo: brain 💩

PR #88, opened 2019-03-05
rewrite table components to use hooks, add pagination

jharting:
I know, I know, this is huge. It is however impossible to switch to hooks without rewriting an entire component :-/


jharting:
@karelhala I am thinking that if the hook definitions here catch on we may eventually move them to frontend components for them to be reusable. Please review especially `src/hooks/table.js`

jharting:
related https://github.com/patternfly/patternfly-react/pull/1503

src/components/RemediationTable.js
```diff
@@ -36,148 +35,164 @@ function formatDate (date) {
...
+                            }]
+                        }
+                        rows={ rows }
+                        { ...sorter.props }
+                        { ...selector.props }
+                    >
+                        <TableHeader/>
+                        <TableBody/>
+                    </Table> :
+                    <p className='ins-c-remediations-table--empty'>No Playbooks found</p>
```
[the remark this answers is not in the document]
jharting: @ryelo good catch! updated

src/hooks/table.js
```diff
@@ -0,0 +1,182 @@
...
+import intersection from 'lodash/intersection';
+
+import { SEARCH_DEBOUNCE_DELAY } from '../constants';
+
+const chain = (f1, f2) => (...args) => {
+    f1(...args);
+    f2(...args);
+};
+
+function callbacks () {
```
karelhala: Phew his function is really hard o understand can you add some comments what does it do?

jharting: @karelhala thanks for feedback. I added a comment

PR #94, opened 2019-03-20
use generated API client

jharting:
https://projects.engineering.redhat.com/browse/RHIPLAT2-445

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1,7 +1,13 @@
 import { API_BASE } from './config';
 
+import axios from 'axios';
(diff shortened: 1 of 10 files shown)


src/api.js
```diff
@@ -1,34 +1,80 @@
...
+    constructor(description) {
+        super('Error communicating with the server');
+        this.description = description;
+    }
 }
 
-// this is here for demo purposes only
-export function getHosts () {
-    return doGet('/api/inventory/v1/hosts');
+async function authInterceptor (config) {
```
karelhala: This is really good, I am thinking if we could move this to FEC so overybody can use these functions in future.

PR #98, opened 2019-03-29
Updating URLs on details page

ryelo:
Updated:
* playbook URL
* system URL should use inventory on current path (insights or rhcs)
* add conditional checking for beta on both

### DIFF
--- a/src/Utilities/urls.js
+++ b/src/Utilities/urls.js
@@ -1,9 +1,35 @@
 import urijs from 'urijs';
 import { getIssuePrefix } from './model';
 
+// Conditionally apply beta
+export function isBeta () {
+
+    const pathName = window.location.pathname.split('/');
+    pathName.shift();
+
+    if (pathName[0] === 'beta') {
+        return ('beta/');
+    }
+
+    return ('');
+}
+
+// Get the current group since we can be mounted at two urls, then add beta conditionally
+export function getGroup () {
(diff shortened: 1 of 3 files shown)


src/Utilities/urls.js
```diff
@@ -28,9 +55,9 @@ export function buildIssueUrl (id) {
 
     switch (getIssuePrefix(id)) {
         case 'advisor':
-            return urijs(document.baseURI).segment('platform').segment('advisor').segment('actions').segment('by_id').segment(parts[1]).toString();
+            return urijs(document.baseURI).segment(`${isBeta()}insights`).segment('advisor').segment('actions').segment('by_id').segment(parts[1]).toString();
```
jharting: this should no longer use the advisor segment, right?

src/Utilities/urls.js
```diff
@@ -1,9 +1,36 @@
...
+    return ('');
+}
+
+// Get the current group since we can be mounted at two urls, then add beta conditionally
+export function getGroup () {
+    const pathName = window.location.pathname.split('/');
+    pathName.shift();
+
+    if (pathName[0] === 'beta') {
+        pathName.shift();
```
jharting: @ryelo this probably should not be here

ryelo: Why not?

ryelo: Oh I see. Removing.

PR #102, opened 2019-04-09
Adds FindItFixIt to Remediations app

benthomasson:
This PR still needs restricting the UI to beta users and possibly to certain user ids.

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -2,6 +2,7 @@ import { Switch, matchPath, Route } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import React from 'react';
 import Home from './routes/Home';
+import FindItFixIt from './routes/FindItFixIt';
 import RemediationDetails from './routes/RemediationDetails';
 
 const InsightsRoute = ({ component: Component, rootClass, ...rest }) => {
@@ -20,7 +21,8 @@ InsightsRoute.propTypes = {
 
 export const routes = {
     home: '/',
-    details: '/:id'
+    details: '/:id',
+    finditfixit: '/fixit'
 };
 
 function checkPaths(group, app) {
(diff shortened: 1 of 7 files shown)


benthomasson:
This still needs more work to integrate with the existing remediations app with the execute playbook button and call the remediations API for the hard-coded values that are in the code currently.

src/Routes.js
```diff
@@ -20,7 +21,8 @@ InsightsRoute.propTypes = {
 
 export const routes = {
     home: '/',
-    details: '/:id'
+    details: '/:id',
+    finditfixit: '/fixit'
```
jharting: I assume eventually there will be a separate page for each playbook and this will be parameterized (e.g. `/:id/fixit`)

src/routes/FindItFixIt.js
```diff
@@ -0,0 +1,275 @@
+/* eslint-disable */
```
jharting: Please do not disable lint

src/routes/logic.js
```diff
@@ -0,0 +1,185 @@
+/* eslint-disable */
```
jharting: This file does not belong to `/routes`. Might be better to move this to a top-level file/folder

src/routes/logic.js
```diff
@@ -0,0 +1,185 @@
...
+      this.tasks_by_host[this.hosts[i].host_id] = [];
+    }
+  }
+
+  send_trace_message(message) {
+    console.log(message)
+  }
+
+	get_api_url(name) {
+    return this.protocol + this.server + '/insights_integration/api/' + name + '/';
```
jharting: `urijs` is already a dependency. You may want to use that to make URL building more robust

src/routes/FindItFixIt.js
benthomasson: Do you have prettier configuration to match your eslint configuration?

src/Routes.js
benthomasson: OK that makes sense.

src/routes/logic.js
benthomasson: Thanks for the tip

src/routes/logic.js
benthomasson: OK

src/routes/FindItFixIt.js
jharting: I don't unfortunately

PR #122, opened 2019-10-10
changed Icon and added 'No' response to replace empty

ryandillinfelton:
Screenshot_20191010_162738
Screenshot_20191010_164155

### DIFF
--- a/src/components/RemediationDetailsTable.js
+++ b/src/components/RemediationDetailsTable.js
@@ -26,7 +26,7 @@ import { useExpander, useFilter, usePagination, useSelector, useSorter } from '.
 import * as debug from '../Utilities/debug';
 
 import './RemediationDetailsTable.scss';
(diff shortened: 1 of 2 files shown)


src/components/RemediationDetailsTable.js
```diff
@@ -53,10 +53,10 @@ function issueDescriptionCell (issue) {
 
 function needsRebootCell (needsReboot) {
     if (needsReboot) {
-        return <CheckIcon/>;
+        return <CheckCircleIcon className="ins-c-remediations-reboot-check-circle"/>;
```
ryelo: can we add an `aria-label="reboot required"` to this? I noticed it wasn't there initially, but this should be super helpful for screen readers.

ryandillinfelton: added it

PR #123, opened 2020-01-08
Add execute button modal

PanSpagetka:
Add execute button modal.
https://projects.engineering.redhat.com/browse/RHCLOUD-3714
To show button you need to type `localStorage.setItem('remediations:fifi:debug', true)` into browser console.

Screenshot from 2020-01-09 10-56-57


src/containers/ExecuteButtons.js
```diff
@@ -6,9 +6,10 @@ import { getConnectionStatus } from '../actions';
 import ExecuteButton from '../components/ExecuteButton';
 
 export const ExecutePlaybookButton = withRouter(connect(
-    ({ connectionStatus: { data, status }}) => ({
+    ({ connectionStatus: { data, status }, selectedRemediation }) => ({
         data,
-        isLoading: status !== 'fulfilled'
+        isLoading: status !== 'fulfilled',
+        issueCount: selectedRemediation.remediation.issues.length
```
RonnyPfannschmidt: seeing this reminds me about a idea we had a while back for better performance by returning the statistics instead of concrete  data from the api to avoid loading the concrete data from the database and passing it around

while this may not manifest before the summit, i believe it might still be helpful to line it up,
as i'm under the impression that with FiFi we will see re-mediations with larger issue counts more often
and its seems like a reasonable opportunity to keep things small/fast

i reckon that we need some more data to give this the right priority

PanSpagetka: I  generally agree with you that it is better to use statistics instead of concrete data. However, this PR adds button that launches modal. At the screen where is the button placed are data in `selectedRemediation.remediation.issues` shown in a table. I don't create new request for issues here, I am just using data we requested earlier.

Even though we already have issue data, it would be better to just use count, unfortunately API doesn't return it. So the issue array is the only thing I can work with now.

RonnyPfannschmidt: i agree, its just something i remembered while noting the data usage
with my current knowledge i believe its not critical to enable yet,
but its worth noting as something that could come up sooner than expected

after data collection i will open up relevant cards

PanSpagetka:
@karelhala review please :)

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,117 @@
...
+
+    const styledConnectionStatus = (status) => (
+        status === 'connected'
+            ? (<div>
+                <CheckCircleIcon
+                    className="ins-c-remediations-reboot-check-circle ins-c-remediations-connection-status"
+                    aria-label="connection status"/>
+                Ready
+            </div>)
+            : status
```
jharting: status is an enum that is not intended to be displayed directly to users. We'll need to somehow "translate" the enum value to a human readable string. The full list of enum values can be found here: https://github.com/RedHatInsights/insights-remediations/blob/master/src/api/openapi.yaml#L925

Some of these human readable labels are listed here: https://marvelapp.com/j3b9b7e/screen/63670941

Here's my suggestion:
* disconnected - Connection issue. Receptor not connected (see mocks)
* no_executor - Cannot remediate - Direct connection. Connect your systems to Satellite to automatically remediate. (see mocks)
* no_source - Cannot remediate - Satellite not configured. Satellite not registered for Playbook execution.
* no_receptor - Cannot remediate - Receptor not configured. Configure Receptor to automatically remediate

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,117 @@
...
+                        key="download"
+                        variant='link' onClick={ () => downloadPlaybook(remediationId) }>
+                        Download Playbook
+                    </Button>
+                ] }
+            >
+                <div>
+                    <TextContent>
+                        <Text component={ TextVariants.p }>
+                            { `Playbook contains ${pluralize(issueCount, 'issue')} affecting ${pluralize(systemCount, 'system')}.` }
```
jharting: This should be the sum of all systems in the Playbook (right now this is sum of systems connected through executors only)

src/components/ExecuteButton.js
karelhala: Good point, I would make this into Object mapper.

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,117 @@
...
+        ([ pass, fail ], e) => (e.connection_status === 'connected' ? [ [ ...pass, e ], fail ] : [ pass, [ ...fail, e ] ])
+        , [ [], [] ]
+    );
+
+    const rows = [ ...connected, ...disconnected ].map(con =>
+        ({ cells: [ con.executor_name || 'Direct connection', con.system_count, { title: styledConnectionStatus(con.connection_status) }]})
+    );
+    const systemCount = connected.reduce((acc, e) => e.system_count + acc, 0);
+
+    const pluralize = (number, str) => number > 1 ? `number ${str}s` : `${number} ${str}`;
```
karelhala: There is typo in the first string literal, instead of `number` wrap it in eval `${number} ${str}s`

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,117 @@
...
+                isOpen={ open }
+                onClose={ () => setOpen(false) }
+                isFooterLeftAligned
+                actions={ [
+                    <Button
+                        key="confirm"
+                        variant="primary"
+                        disabled={ connected.length > 0 }
+                        onClick={ () => setOpen(false) }>
+                        { `Execute Playbook on ${systemCount} items` }
```
karelhala: You should use the `pluralize` function in here.

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,183 @@
...
+                isOpen={ open }
+                onClose={ () => setOpen(false) }
+                isFooterLeftAligned
+                actions={ [
+                    <Button
+                        key="confirm"
+                        variant="primary"
+                        disabled={ connected.length > 0 }
+                        onClick={ () => setOpen(false) }>
+                        { `Execute Playbook on ${pluralize(connectedCount, 'system')}` }
```
ryelo: according to the mockups, `Playbook` should be lower case

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,183 @@
...
+                    </Button>,
+                    <Button
+                        key="download"
+                        variant='link' onClick={ () => downloadPlaybook(remediationId) }>
+                        Download Playbook
```
ryelo: according to the mockups, `Playbook` should be lower case

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,183 @@
...
+                    </Button>
+                ] }
+            >
+                <div>
+                    <TextContent>
+                        <Text component={ TextVariants.p }>
+                            { `Playbook contains ${pluralize(issueCount, 'issue')} affecting ${pluralize(systemCount, 'system')}.` }
```
ryelo: `${pluralize(systemCount, 'system')}` should be bold

ryelo: `${pluralize(issueCount, 'issue')}` should be bold

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,183 @@
...
+                        </Text>
+                        <Text component={ TextVariants.p }>
+                        Systems connected to a Satelite instance and configured with Receptor can be automatically remediated.
```
ryelo: `Satelite` should be `Satellite`

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,183 @@
...
+                            To remediate other systems, download the Ansible playbook for this plan.
+
+                        </Text>
+                        <Text component={ TextVariants.h2 }>Connection status of systems</Text>
```
ryelo: I think this should be an `h3` or `h4` from looking at the mockups. It should be similar in size to the text above it, but bold.

src/components/ExecuteButton.js
jharting: The mockups are wrong. We use Playbook everywhere because it is sort of Ansible's trademark thing.

src/components/ExecuteButton.js
jharting: Same as above

src/components/ExecuteButton.js
ryelo: In that case, do we want to capitalize `playbook` in the modal header?

jharting: Yes, good catch!

ryelo: Going to update my review

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,183 @@
...
+    );
+    const connectedCount = connected.reduce((acc, e) => e.system_count + acc, 0);
+    const systemCount = data.reduce((acc, e) => e.system_count + acc, 0);
+
+    const pluralize = (number, str) => number > 1 ? `${number} ${str}s` : `${number} ${str}`;
+    return (isUserEntitled
+        ?  <React.Fragment>
+            <Button
+                onClick={ () => { setOpen(true); getConnectionStatus(remediationId); } }>
+        Execute playbook
```
ryelo: `playbook` should be capitalized

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,183 @@
...
+            </Button>
+            <Modal
+                className="ins-c-dialog"
+                width={ '50%' }
+                title={ 'Execute playbook' }
```
ryelo: `playbook` should be capitalized

src/components/ExecuteButton.js
```diff
@@ -0,0 +1,183 @@
...
+                ] }
+            >
+                <div>
+                    <TextContent>
+                        <Text component={ TextVariants.p }>
+                            { `Playbook contains ${pluralize(issueCount, 'issue')} affecting ${pluralize(systemCount, 'system')}.` }
+                        </Text>
+                        <Text component={ TextVariants.p }>
+                        Systems connected to a Satelite instance and configured with Receptor can be automatically remediated.
+                            To remediate other systems, download the Ansible playbook for this plan.
```
ryelo: `playbook` should be capitalized

src/components/ExecuteButton.js
PanSpagetka: Are you sure? I know that `Download Playbook` is using upper case P, but mock are consistently using lower case p.

jharting: Yes, in the past the Ansible team asked us to always use capitalized "Playbook" everywhere.

src/components/ExecuteButton.scss
```diff
@@ -0,0 +1,7 @@
+.ins-c-remediations-connection-status {
+  margin-right: 5px;
```
karelhala: Use `var(--pf-global--spacer--xs)` instead of direct number, also align this to one line.

src/components/ExecuteButton.scss
```diff
@@ -0,0 +1,7 @@
...
+}
+
+.ins-c-remediations-connection-status-error {
+  color: #E10000;
```
karelhala: Use `var(--pf-chart-global--danger--Color--100)` instead of direct hex color, also align this to one line.

src/components/ExecuteButton.js
katierik: Our capitalization guide had been recommending lowercase, but I've updated based on this info that it should be uppercase. 
Mocks may be inconsistent while I work on getting them updated

ryelo:
I think we should just update the (custom config)[https://github.com/RedHatInsights/insights-remediations-frontend/blob/master/.travis/custom_release.sh] so that master pushes to ci/qa-beta and *not* prod-beta

jharting:
@ryelo nah let's hide it under that localStorage thing. That way we can start shifting our mindset towards feature flags and CD :wink:

PR #124, opened 2020-01-15
Systems for action modal

PanSpagetka:
Add new modal that displays system for remediation.
We don't have data to for `Connection type` column, so we can't filter a group by type yet.

https://projects.engineering.redhat.com/browse/RHCLOUD-3622

Screenshot from 2020-01-15 14-28-22

Screenshot from 2020-01-15 14-28-38

Screenshot from 2020-01-15 14-28-50

### DIFF
--- a/src/components/SystemForActionButton.js
+++ b/src/components/SystemForActionButton.js
@@ -4,22 +4,32 @@ import * as pfReactTable from '@patternfly/react-table';
 import * as reactCore from '@patternfly/react-core';
 import * as reactIcons from '@patternfly/react-icons';
 import * as reactRouterDom from 'react-router-dom';
+import { connect } from 'react-redux';
 
 import PropTypes from 'prop-types';
 import {
-    Button, Modal, TextContent, Text, TextVariants
+    Button, Modal
 } from '@patternfly/react-core';
 import { getRegistry } from '@redhat-cloud-services/frontend-components-utilities/files/Registry';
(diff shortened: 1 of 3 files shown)


src/components/SystemForActionButton.js
```diff
@@ -0,0 +1,133 @@
...
+    return (
+        <React.Fragment>
+            <Button
+                variant='link' onClick={ () => setOpen(true) }>
+                { issue.systems.length }
+            </Button>
+            <Modal
+                className="ins-c-dialog"
+                width={ '50%' }
+                title={ `System for action ${issue.description}` }
```
karelhala: This should be plural, if more than 1 system is in action.

src/components/SystemForActionButton.js
```diff
@@ -0,0 +1,133 @@
...
+                className="ins-c-dialog"
+                width={ '50%' }
+                title={ `System${issue.systems.length > 1 ? 's' : ''} for action ${issue.description}` }
+                isOpen={ open }
+                onClose={ () => setOpen(false) }
+                isFooterLeftAligned
+            >
+                { InventoryTable && <InventoryTable
+                    ref={ inventory }
+                    items={ issue.systems }
```
jharting: These should be ordered the same way they were in the old table

jharting: The system name should appear as a hyperlink (i.e. blue) ideally opening a new tab when clicked

Screenshot_20200129_081714

PR #125, opened 2020-01-17
Bind endpoint to Execute Playbook button

PanSpagetka:
Bind endpoint to Execute Playbook button.

To show `Reset etag` button you should type `localStorage.setItem('remediations:fifi:debug', true)` into browser console.

Screenshot from 2020-01-27 13-51-17
Screenshot from 2020-01-28 10-47-14

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -74,3 +74,8 @@ export const runRemediation = (id, etag) => {
     };
 };
 
+export const setEtag = (etag) => ({
+    type: ACTION_TYPES.SET_ETAG,
+    payload: { etag }
+});
+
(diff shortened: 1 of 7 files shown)


src/actions.js
```diff
@@ -62,3 +62,11 @@ export const getConnectionStatus = (id) => {
         payload: remediations.getRemediationConnectionStatus(id)
     };
 };
+
+export const runRemediation = (id, etag) => {
+    return {
+        type: ACTION_TYPES.RUN_REMEDIATION,
+        payload: remediations.runRemediation(id, { headers: { 'If-None-Match': etag }})
```
jharting: this should be `if-match` (the problem has been fixed)

PanSpagetka:
@jharting I have added modal refresh and info text when 412. Rest button is displayed when `remediations:fifi:debug` is set, same as the `Execute Playbook` button on remediation detail page.

One thing that we might want to address is what happens  when other error happened. Right now modal stays open and notification is displayed, but that is more likely UX question.

@katierik what do you think?

PR #127, opened 2020-01-20
Add inline alert

ryelo:
https://marvelapp.com/j3b9b7e/screen/63670935

This adds the inline alert for `Automatic remediation in Progress` and the toast notification for the status of the remediation

Note: This does *not* hook them up to the correct actions, just adds them as components.

I'm using `mount` in the hooks test because there is a bug with shallow rendering and hooks


src/components/Alerts/RemediationStatusToast.js
```diff
@@ -0,0 +1,38 @@
+import React, { useState, useEffect } from 'react';
+import propTypes from 'prop-types';
+
+import { Alert } from '@patternfly/react-core';
+import './RemediationStatusToast.scss';
+
+function RemediationStatusToast ({ status, name, ...props }) {
```
karelhala: Shouldn't we use for the Toast notifications use FEC notifications?

ryelo: Yeah we definitely can go down that route. If that's easier, I can remove the toast from this PR.

PR #128, opened 2020-01-24
Execute alert banner

tahmidefaz:
Jira: [Here](https://projects.engineering.redhat.com/browse/RHCLOUD-3715)

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -63,8 +63,6 @@ export const getConnectionStatus = (id) => {
     };
 };
 
-export const toggleExecutePlaybookBanner = () => {
-    return {
-        type: ACTION_TYPES.EXECUTE_PLAYBOOK_BANNER
-    };
-};
+export const toggleExecutePlaybookBanner = () => ({
+    type: ACTION_TYPES.EXECUTE_PLAYBOOK_BANNER
+});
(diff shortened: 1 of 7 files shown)


src/actions.js
```diff
@@ -62,3 +62,9 @@ export const getConnectionStatus = (id) => {
         payload: remediations.getRemediationConnectionStatus(id)
     };
 };
+
+export const toggleExecutePlaybookBanner = () => {
+    return {
```
karelhala: No need to have return statement here.

src/components/ExecuteButton.js
```diff
@@ -123,7 +123,7 @@ const ExecuteButton = ({ isLoading, data, getConnectionStatus, remediationId, is
                         key="confirm"
                         variant="primary"
                         isDisabled={ connected.length === 0 }
-                        onClick={ () => setOpen(false) }>
+                        onClick={ () => { setOpen(false); toggleExecutePlaybookBanner(); } }>
```
karelhala: Can you span it to 2 lines so it is easier to read?

src/routes/RemediationDetails.js
```diff
@@ -72,6 +74,10 @@ class RemediationDetails extends Component {
 
         return (
             <React.Fragment>
+                {
+                    getStore().getState().executePlaybookBanner.isVisible &&
```
karelhala: This should come from props, since it is in connected function.

PR #129, opened 2020-01-27
RHCLOUD-4230 Patch Manager support in frontend

jharting:
https://projects.engineering.redhat.com/browse/RHCLOUD-4230


src/Utilities/model.js
```diff
@@ -11,6 +11,7 @@ export function getIssueApplication ({ id }) {
         case 'advisor': return 'Insights';
         case 'ssg': return 'Compliance';
         case 'vulnerabilities': return 'Vulnerability';
+        case 'patch-advisory': return 'Patch';
```
RonnyPfannschmidt: it seems to me like it might be a nice follow-up to turn those growing case expressions into constant mappings

PR #130, opened 2020-01-27
Start logic for toast notification on execute

ryelo:


src/routes/RemediationDetails.js
```diff
@@ -47,6 +48,26 @@ class RemediationDetails extends Component {
         this.props.switchAutoReboot(this.id, autoReboot);
     };
 
+    handleSuccessBanner = (id, name) => {
```
karelhala: This function is not called anywhere, is that correct?

src/routes/RemediationDetails.js
```diff
@@ -47,6 +48,26 @@ class RemediationDetails extends Component {
...
+            title: `Remediation plan ${name} successfully completed.`,
+            dismissDelay: 8000
+        });
+    }
+
+    handlePlaybookCancel = (id, name) => {
+        // TODO: Cancel playbook
+        this.props.toggleExecutePlaybookBanner();
+        this.props.addNotification({
+            variant: 'danger',
```
karelhala: Is danger correct in here? Since this will be shown everytime user closes playbook execution I would make this warning instead of danger.

src/routes/RemediationDetails.js
ryelo: At the moment, no it's not. But when the API is hooked up, this should be the function that is called when the playbook has finished.

src/routes/RemediationDetails.js
ryelo: Yeah, this is when a user cancels a playbook. This wasn't in the mocks, but it's something that we want. I'll talk to UXD to figure it out!

src/routes/RemediationDetails.js
karelhala: Nice planning ahead!

src/routes/RemediationDetails.js
ryelo: >>> 
Katie Riker  2 minutes ago
good idea. Should be "info" toast (not a success, not a warning, not an error).
With a 2 second think: "Canceled execution of playbook [foo]" -- but let me circle back on this. I'll add to mocks

PR #131, opened 2020-01-28
Plan summary card

ryelo:
Screen Shot 2020-02-11 at 11 29 36 AM
Screen Shot 2020-02-11 at 11 02 31 AM
Screen Shot 2020-02-11 at 11 02 28 AM

### DIFF
--- a/src/components/Layouts/DescriptionList.js
+++ b/src/components/Layouts/DescriptionList.js
@@ -4,15 +4,20 @@ import classnames from 'classnames';
 
 import './DescriptionList.scss';
 
-const DescriptionList = ({ title, children, isBold, ...props }) => {
+const DescriptionList = ({ title, className, children, isBold, ...props }) => {
 
     const DescriptionListClasses = classnames(
         'ins-l-description-list__description',
         { ['ins-l-description-list__description--bold']: isBold }
     );
 
+    const mainClasses = classnames(
+        className,
+        'ins-l-description-list'
+    );
+
     return (
-        <dl className='ins-l-description-list' { ...props }>
+        <dl className={ mainClasses } { ...props }>
             <dt className='ins-l-description-list__title'><b>{ title }</b></dt>
             <dd className={ DescriptionListClasses }> { children } </dd>
         </dl>
@@ -24,5 +29,6 @@ export default DescriptionList;
 DescriptionList.propTypes = {
     title: propTypes.string,
     children: propTypes.any,
-    isBold: propTypes.bool
+    isBold: propTypes.bool,
+    className: propTypes.string
 };
--- a/src/components/Layouts/DescriptionList.js
+++ b/src/components/Layouts/DescriptionList.js
@@ -6,7 +6,7 @@ import './DescriptionList.scss';
 
 const DescriptionList = ({ title, className, children, isBold, ...props }) => {
 
-    const DescriptionListClasses = classnames(
+    const descriptionListClasses = classnames(
         'ins-l-description-list__description',
         { ['ins-l-description-list__description--bold']: isBold }
     );
@@ -19,7 +19,7 @@ const DescriptionList = ({ title, className, children, isBold, ...props }) => {
     return (
         <dl className={ mainClasses } { ...props }>
             <dt className='ins-l-description-list__title'><b>{ title }</b></dt>
-            <dd className={ DescriptionListClasses }> { children } </dd>
+            <dd className={ descriptionListClasses }> { children } </dd>
         </dl>
     );
 };
--- a/src/components/Layouts/DescriptionList.js
+++ b/src/components/Layouts/DescriptionList.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import classnames from 'classnames';
+
+import './DescriptionList.scss';
+
+const DescriptionList = ({ title, children, isBold, ...props }) => {
+
+    const DescriptionListClasses = classnames(
+        'ins-l-description-list__description',
+        { ['ins-l-description-list__description--bold']: isBold }
+    );
+
+    return (
+        <dl className='ins-l-description-list' { ...props }>
+            <dt className='ins-l-description-list__title'><b>{ title }</b></dt>
+            <dd className={ DescriptionListClasses }> { children } </dd>
+        </dl>
+    );
+};
+
+export default DescriptionList;
+
+DescriptionList.propTypes = {
+    title: propTypes.string,
+    children: propTypes.any,
+    isBold: propTypes.bool
+};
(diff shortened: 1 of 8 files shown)


src/components/Layouts/DescriptionList.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import classnames from 'classnames';
+
+import './DescriptionList.scss';
+
+const DescriptionList = ({ title, className, children, isBold, ...props }) => {
+
+    const DescriptionListClasses = classnames(
```
karelhala: Can you make it lowercase `description`?

src/components/Layouts/DescriptionList.scss
```diff
@@ -0,0 +1,13 @@
+.ins-l-description-list {
+    .ins-l-description-list__title {
+        margin-bottom: 6px; // form-element-spacer
+        font-size: var(--pf-global--FontSize--md);
+    }
+    .ins-l-description-list__description {
+        font-size: var(--pf-global--FontSize--md);
+        &--bold {
+            font-size: var(--pf-global--FontSize--lg);
+            font-weight: 700;
```
karelhala: Isn't there pf var for font weight?

ryelo:
@karelhala There were mixed use of single and double quotes in this file. I just switched them all to single, but I can revert that and have them mixed in.

jharting:
hey @ryelo pls check the notes on terminology in https://projects.engineering.redhat.com/browse/RHCLOUD-4177

ryelo:
@jharting Fixed. I totally didn't read that Jira card very well (obviously). Screenshot has been updated too.

src/routes/RemediationDetails.js
```diff
@@ -175,6 +181,37 @@ class RemediationDetails extends Component {
...
                         </StackItem>
+                        { localStorage.getItem('remediations:fifi:debug') === 'true' &&
+                            <StackItem>
+                                <Card>
+                                    <CardHeader className='ins-m-card__header-bold'>Playbook Summary</CardHeader>
+                                    <CardBody>
+                                        <Flex className='ins-c-playbookSummary' breakpointMods={ [{ modifier: FlexModifiers.column }] }>
+                                            <Flex className='ins-c-playbookSummary__overview'>
+                                                <FlexItem breakpointMods={ [{ modifier: FlexModifiers['spacer-xl'] }] }>
+                                                    <DescriptionList isBold title='Total systems'>{ totalSystems } systems</DescriptionList>
```
karelhala: You can use the pluralize function here as well.

src/routes/RemediationDetails.js
```diff
@@ -175,6 +181,41 @@ class RemediationDetails extends Component {
...
+                                                    </DescriptionList>
+                                                </FlexItem>
+                                            </Flex>
+                                            <DescriptionList className='ins-c-playbookSummary__settings' title='Playbook settings'>
+                                                <Flex>
+                                                    <FlexItem breakpointMods={ [{ modifier: FlexModifiers['spacer-xl'] }] }>
+                                                        Autoreboot:&nbsp;
+                                                        <b> { remediation.auto_reboot && remediation.needs_reboot ? 'Enabled' : 'Disabled' } </b>
+                                                    </FlexItem>
+                                                    <FlexItem>{ pluralize(stats.systemsWithReboot, 'system') } require reboot</FlexItem>
```
jharting: should "require" also be pluralized?

src/routes/RemediationDetails.js
```diff
@@ -175,6 +181,41 @@ class RemediationDetails extends Component {
                                 </GridItem>
                             </Grid>
                         </StackItem>
+                        { localStorage.getItem('remediations:fifi:debug') === 'true' &&
```
jharting: I don't think we need to guard this with a "feature flag". I think this change in mockups was intended to be a "lipstick" improvement, replacing the existing "System reboot" and "Playbook details" cards. The new card has nothing specific to FiFi in it so we can just release it.

ryelo: Cool. I'll remove the older bits in this PR too.

src/routes/RemediationDetails.js
ryelo: `system(s)` should be plural in front of `require` so it does not need to be pluralized.

jharting: I meant the case when you only have 1 system - it should say require**s**. Sorry for confusion.

src/routes/RemediationDetails.js
```diff
@@ -131,70 +133,39 @@ class RemediationDetails extends Component {
...
+                                                    title='Total systems'>
+                                                    { pluralize(totalSystems, 'system') }
+                                                </DescriptionList>
+                                            </FlexItem>
+                                        </Flex>
+                                        <DescriptionList className='ins-c-playbookSummary__settings' title='Playbook settings'>
+                                            <Flex>
+                                                <FlexItem breakpointMods={ [{ modifier: FlexModifiers['spacer-xl'] }] }>
+                                                    Autoreboot:&nbsp;
+                                                    <b> { remediation.auto_reboot && remediation.needs_reboot ? 'Enabled' : 'Disabled' } </b>
```
jharting: In the mocks this is dark green then "Enabled". It is not clear what color the disabled option should be. Also, when the Playbook does not require reboot at all we should probably call it out explicitly rather than say "Disabled". Let me get that clarified with UXD

src/routes/RemediationDetails.js
```diff
@@ -131,70 +133,39 @@ class RemediationDetails extends Component {
...
+                                                </FlexItem>
+                                                <FlexItem>{ pluralize(stats.systemsWithReboot, 'system') } require reboot</FlexItem>
```
jharting: what was the conclusion on require vs requires? Do we not pluralize to that level of detail or should this be fixed?

src/routes/RemediationDetails.js
ryelo: Ah yes you're right then.

src/routes/RemediationDetails.js
ryelo: It's being fixed. Thanks for finding it!

src/routes/RemediationDetails.js
ryelo: Sounds good! I'll leave this conversation open, but I'm going to push up a change for the "require(s)"

ryelo:
Screen Shot 2020-02-11 at 11 29 36 AM
Screen Shot 2020-02-11 at 11 02 31 AM
Screen Shot 2020-02-11 at 11 02 28 AM

karelhala:
Codewise looks good @jharting what are your thoughts on the content? Safe to merge?

jharting:
Looks great!

PR #135, opened 2020-02-25
Implement RBAC in the UI

ryelo:
## What
implement RBAC roles in the UI

## Jira
https://projects.engineering.redhat.com/browse/RHCLOUD-4500

### No perms, but entitled
Screen Shot 2020-02-27 at 3 17 12 PM

### Read
Screen Shot 2020-02-27 at 1 59 33 PM
Screen Shot 2020-02-27 at 1 59 40 PM

### Read+ Write
Screen Shot 2020-02-27 at 1 56 59 PM
Screen Shot 2020-02-27 at 1 57 06 PM

### Read + Write + Execute
Screen Shot 2020-02-27 at 1 58 40 PM
Screen Shot 2020-02-27 at 1 58 48 PM

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -17,7 +17,8 @@ class App extends Component {
         this.state = {
             readPermission: false,
             writePermission: false,
-            executePermission: false
+            executePermission: false,
+            isOrgAdmin: false
         };
     }
 
@@ -30,8 +31,17 @@ class App extends Component {
     componentDidMount () {
         insights.chrome.init();
         insights.chrome.identifyApp('remediations');
-        // TODO: Do the user check and set this function below to the right values
-        this.handlePermissionUpdate(true, false, false);
+        window.insights.chrome.getUserPermissions().then(
+            allPermissions => {
+                const permissionList = allPermissions.map(permissions => permissions.permission);
+                this.handlePermissionUpdate(
+                    permissionList.includes('remediations:remediation:read'),
+                    permissionList.includes('remediations:remediation:write'),
+                    permissionList.includes('remediations:remediation:execute')
+                );
+            }
+        );
+        window.insights.chrome.auth.getUser().then((user) => user.identity.user.is_org_admin && this.setState({ isOrgAdmin: true }));
     }
 
     componentWillUnmount () {
@@ -43,9 +53,11 @@ class App extends Component {
         return (
             <PermissionContext.Provider
                 value={ {
-                    readPermission: this.state.readPermission,
-                    writePermission: this.state.writePermission,
-                    executePermission: this.state.executePermission
+                    permissions: {
+                        read: this.state.isOrgAdmin ? true : this.state.readPermission,
+                        write: this.state.isOrgAdmin ? true : this.state.writePermission,
+                        execute: this.state.isOrgAdmin ? true : this.state.executePermission
+                    }
                 } }>
                 <NotificationsPortal />
                 <Routes childProps={ this.props } />
--- a/src/App.js
+++ b/src/App.js
@@ -17,8 +17,7 @@ class App extends Component {
         this.state = {
             readPermission: undefined,
             writePermission: undefined,
-            executePermission: undefined,
-            isOrgAdmin: undefined
+            executePermission: undefined
         };
     }
 
@@ -41,7 +40,6 @@ class App extends Component {
                 );
             }
         );
-        window.insights.chrome.auth.getUser().then((user) => user.identity.user.is_org_admin && this.setState({ isOrgAdmin: true }));
     }
 
     componentWillUnmount () {
@@ -54,9 +52,9 @@ class App extends Component {
             <PermissionContext.Provider
                 value={ {
                     permissions: {
-                        read: this.state.isOrgAdmin || this.state.readPermission,
-                        write: this.state.isOrgAdmin || this.state.writePermission,
-                        execute: this.state.isOrgAdmin || this.state.executePermission
+                        read: this.state.readPermission,
+                        write: this.state.writePermission,
+                        execute: this.state.executePermission
                     }
                 } }>
                 <NotificationsPortal />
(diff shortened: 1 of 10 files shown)


ryelo:
Make sure to turn on the fifi flag to see the execute button 😄

src/App.js
```diff
@@ -8,11 +8,40 @@ import './App.scss';
...
+            allPermissions => {
+                const permissionList = allPermissions.map(permissions => permissions.permission);
+                this.handlePermissionUpdate(
+                    permissionList.includes('remediations:remediation:read'),
+                    permissionList.includes('remediations:remediation:write'),
+                    permissionList.includes('remediations:remediation:execute')
+                );
+            }
+        );
+        window.insights.chrome.auth.getUser().then((user) => user.identity.user.is_org_admin && this.setState({ isOrgAdmin: true }));
```
jharting: This calback may complete at different time than the permission one. I.e. an org admin without permissions would see "you cannot use this" blip first and then it would work. Instead I think we should wait for both these calls to finish before dismissing the loading screen

ryelo:
@jharting 
* I reworked some of the consumers to use the hook implementation of context.
* removed the "remove action" from the skeleton. I'd say we remove it for now, but if it's ever decided to just disable buttons instead of hiding them, we can add it back and just keep it disabled
* I removed the org admin check
* removed the checkboxes if you don't have write perms on the details page

src/App.js
```diff
@@ -8,11 +8,38 @@ import './App.scss';
...
+    handlePermissionUpdate = (hasRead, hasWrite, hasExecute) => this.setState({
+        readPermission: hasRead,
+        writePermission: hasWrite,
+        executePermission: hasExecute
+    });
+
     componentDidMount () {
         insights.chrome.init();
         insights.chrome.identifyApp('remediations');
+        window.insights.chrome.getUserPermissions().then(
```
jharting: This fetches all permissions which is wasteful. It should be enough to pass "remediations" to the function

jharting: However, even when I change the code to pass the app name it still does not seem to be making the correct REST request :-/

src/App.js
```diff
@@ -8,11 +8,38 @@ import './App.scss';
...
+            allPermissions => {
```
jharting: Should we delay loading of data (i.e. keep a spinner on) until after this finishes? Otherwise a data fetch request will cause 403 and an error toast in addition to the DeniedState

src/App.js
ryelo: This is currently calling the chrome function. I don't believe we have a way to filter out a certain app at this time.

src/App.js
```diff
@@ -8,11 +8,38 @@ import './App.scss';
...
+                const permissionList = allPermissions.map(permissions => permissions.permission);
+                this.handlePermissionUpdate(
+                    permissionList.includes('remediations:remediation:read'),
```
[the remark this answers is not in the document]
ryelo: Are you saying that if a user has any `remediations:foo:bar` permission, they also get a "read" permission?

src/App.js
jharting: @ryelo isn't it what this first param does? https://github.com/RedHatInsights/insights-chrome/blob/master/src/js/entry.js#L125

jharting: Or rather supposed to do? Like I said when testing it did not seem to work

src/App.js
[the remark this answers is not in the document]
jharting: @ryelo Not at all. Each permission has three segments separated by colons i.e. application:resource_type:operation. Each of these may either have a constant value or asterisk may be used as a wildcard. If an orgadmin gives somebody a `remediations:*:*` permission it means they can do whatever in Remediations. The orgadmin may decide to assign explicit constant permissions (e.g. `remediations:remediation:read` and `remediations:remediation:write`), use a wildcard (e.g. `remediations:remediation:*`) or use a combination. Therefore, we should enable "write" controls not only in cases where we discover user having the constant write permission (`remediations:remediation:write`) but also look for wildcard ones that match.

See https://docs.google.com/document/d/1x1kbj7YvEz4IOWEp2-wis2OlvUCcVwkcc4aTPsfRIAU/edit# for more details.

src/App.js
karelhala: It is exactly what @jharting is refering to, it is strange though because I tested it and it does exactly what was expected. It queries the RBAC server with applications parameter in query URL

ryelo: This is working for me! I'll update accordingly

src/routes/Home.js
```diff
@@ -69,24 +72,32 @@ class Home extends Component {
...
-                    title="Create Plan"
-                    className='ins-c-plan-modal'
-                    onClose = { this.onClose }
-                    isOpen= { isModalOpen }
-                    content = { ModalStepContent }
-                />
-            </React.Fragment>
+            <PermissionContext.Consumer>
+                { value =>
+                    value.permissions.read === false &&
```
jharting: This should only depend on the read permission - i.e. if the user does not have read permission they won't see anything (as the API won't return data without the read permission)

src/routes/RemediationDetails.js
```diff
@@ -143,11 +145,15 @@ class RemediationDetails extends Component {
                         <LevelItem>
                             <Split gutter="md">
                                 { this.state.isEntitled &&
-                                    <SplitItem>
-                                        <ExecutePlaybookButton
-                                            remediationId={ remediation.id }>
-                                        </ExecutePlaybookButton>
-                                    </SplitItem>
+                                    <PermissionContext.Consumer>
```
jharting: Can we add the `<DeniedState/>` logic all routes (meaning here)? I.e. if the read permission is taken away from me `/remediations` gets me to `<DeniedState/>` but so should `/remediations/12345` (e.g. an old bookmark)

PR #141, opened 2020-03-30
Fix debug remediations button

karelhala:


jharting:
@karelhala nevermind, I am dumb

PR #143, opened 2020-04-01
playbook activity

jharting:


src/components/statusHelper.js
```diff
@@ -0,0 +1,102 @@
...
+                { renderStatus('success', `${systemsStatus.success}`) }
+            </FlexItem>
+            <FlexItem>
+                { renderStatus('failure', `${systemsStatus.failure}`) }
+            </FlexItem>
+            <FlexItem>
+                { renderStatus('running', `${systemsStatus.running}`) }
+            </FlexItem>
+            
+            { permission.permissions.execute && systemsStatus.running > 0 &&
```
jharting, line 69: Instead of `systemsStatus.running` this should be based on executor.status i.e. the button should show up if executor.status is one of [running, pending, acked]

src/components/statusHelper.js
```diff
@@ -0,0 +1,102 @@
...
+            }
+        </Flex>
+    );
+
+    if (needsTooltip) {
+        return <Tooltip
+            position='right'
+            enableFlip
+            content={
+                <span>{ `Run: ${statusTextPlain(executorStatus)} |
```
jharting, line 82: The tooltip should show the non-normalized statuses

src/routes/RemediationDetails.js
```diff
@@ -169,6 +209,9 @@ class RemediationDetails extends Component {
                                                                 { pluralize(totalSystems, 'system') }
                                                             </DescriptionList>
                                                         </FlexItem>
+                                                        { playbookRuns &&
+                                                            this.renderLatestActivity(playbookRuns)
```
jharting, line 213: Can we make this so that once the user submits the Execute button modal this part of the page auto-refreshes so that the Latest activity (which is going to be the one the user just submitted) shows up here without the user having to refresh the page?

jharting: Also, when the latest activity is in the running state can we re-fetch the list of playbook runs (e.g. every 30 seconds) for the Latest activity status to auto-update?

src/routes/RemediationDetails.js
```diff
@@ -92,8 +107,32 @@ class RemediationDetails extends Component {
         return (status ? 'Enabled' : 'Disabled');
     }
 
+    renderLatestActivity = (playbookRuns) => {
+
+        if (playbookRuns.length) {
+
+            const mostRecentDate = Math.max(...playbookRuns.map(playbooks => new Date(playbooks.updated_at)), 0);
```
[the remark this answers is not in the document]
ryelo: oh that's nice to know 😄

src/routes/RemediationDetails.js
```diff
@@ -92,8 +107,32 @@ class RemediationDetails extends Component {
...
+            const mostRecentPlaybook = playbookRuns.find(allPlaybookRuns => new Date(allPlaybookRuns.updated_at).getTime() === mostRecentDate);
+
+            return (
+                <FlexItem breakpointMods={ [{ modifier: FlexModifiers['spacer-xl'] }] }>
+                    <DescriptionList
+                        hasGutter
+                        title='Latest activity'>
+                        { renderStatusIcon(mostRecentPlaybook.status) }
+                        <span><DateFormat type='exact' date={ mostRecentPlaybook.updated_at } /></span>
```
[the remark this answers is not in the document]
PanSpagetka: After updating FE components version, it should be consistent.

src/components/statusHelper.js
PanSpagetka: Does it mean that we should change `nomalizeStatus` function?
to:
```
export const normalizeStatus = (status) => ({
    running: 'running',
    pending: 'running',
    failure: 'failure',
    canceled: 'failure',
    success: 'success',
    acked: 'running'
})[status];
```

src/routes/RemediationDetails.js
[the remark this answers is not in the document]
ryelo: This should be solved with an update to FEC

src/components/RemediationActivityTable.js
```diff
@@ -0,0 +1,89 @@
...
+                    fullWidth: true,
+                    cells: [{
+                        title: <Table
+                            aria-label="Compact expandable table"
+                            cells={ [ 'Connection', 'Systems', 'Playbook runs status' ] }
+                            rows={ playbooks.executors.map(e => (
+                                { cells: [
+                                    { title: <Link to={ `/${remediation.id}/${playbooks.id}/${e.executor_id}` }>{ e.executor_name }</Link> },
+                                    e.system_count,
+                                    { title: statusSummary(normalizeStatus(playbooks.status), systemsStatus, permission, true) }
```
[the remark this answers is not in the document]
ryelo: Going to make this a prop for StatusSummary and not show it if there is no `onClick`

src/routes/RemediationDetails.js
```diff
@@ -209,8 +251,20 @@ class RemediationDetails extends Component {
...
+                                    <StackItem className='ins-c-playbookSummary__tabs'>
+                                        <Tabs activeKey={ this.state.activeTabKey } onSelect={ this.handleTabClick }>
+                                            <Tab eventKey={ 0 } title='Issues'>
+                                                <RemediationDetailsTable remediation={ remediation } status={ this.props.selectedRemediationStatus }/>
+                                            </Tab>
+                                            <Tab eventKey={ 1 } title='Activity'>
+                                                { this.state.isEntitled
+                                                    ? playbookRuns && playbookRuns.length
+                                                        ? <RemediationActivityTable remediation={ remediation } playbookRuns={ playbookRuns }/>
+                                                        : <NotConfigured/>
```
jharting, line 263: If the playbookRuns list is empty we should show this: https://marvelapp.com/j3b9b7e/screen/63670938

`<NotConfigured/>` should only be shown when receptor is not configured. Whether receptor is configured or not can be figured out from an API call to sources API. See https://projects.engineering.redhat.com/browse/RHCLOUD-5402 for details

src/routes/RemediationDetails.js
```diff
@@ -74,10 +90,15 @@ class RemediationDetails extends Component {
 
         const { entitlements } = await window.insights.chrome.auth.getUser();
 
+        this.loadPlaybookRuns();
+
         this.setState({
             isEntitled: entitlements.smart_management.is_entitled
         });
 
+        fetch(`${window.location.origin}/api/sources/v2.0/endpoints?filter[receptor_node][not_nil]`)
```
jharting, line 99: Can we put this to App.js so that we do not flood sources API with request on each page load (similarly to RBAC)? Also, might be better to put the actual fetch function call together with the URL definition to api.js and import it from there.

src/routes/RemediationDetails.js
```diff
@@ -92,8 +113,43 @@ class RemediationDetails extends Component {
...
+        if (playbookRuns.length) {
+
+            const mostRecent = playbookRuns[0];
+
+            return (
+                <FlexItem breakpointMods={ [{ modifier: FlexModifiers['spacer-xl'] }] }>
+                    <DescriptionList
+                        hasGutter
+                        title='Latest activity'>
+                        { renderStatusIcon(mostRecent.status) }
```
jharting, line 127: This will receive one of these https://github.com/RedHatInsights/insights-remediations/blob/master/src/api/openapi.yaml#L623

Please update this for the icon to be returned based on a normalized status. Also, can we wrap the icon with a tooltip that shows non-normalized status text? E.g. if the playbook_run is in `pending` state you would see "running" icon with a tooltip text that clarifies is it currently pending?

src/components/RemediationActivityTable.js
```diff
@@ -0,0 +1,109 @@
...
+                    isOpen: false,
+                    cells: [
+                        { title: <Link to={ `/${remediation.id}/${playbooks.id}` }><DateFormat type='exact' date={ playbooks.created_at }/></Link>,
+                            cellFormatters: [ expandable ]},
+                        `${playbooks.created_by.first_name} ${playbooks.created_by.last_name}`,
+                        { title: <StatusSummary
+                            executorStatus={ normalizeStatus(playbooks.status) }
+                            counts={ playbooks.executors.reduce((acc, ex) => (
+                                { pending: acc.pending + ex.counts.pending,
+                                    running: acc.running + ex.counts.running,
```
jharting, line 37: Acked is missing. See all the possible statuses for an executor here: https://github.com/RedHatInsights/insights-remediations/blob/master/src/api/openapi.yaml#L625

src/components/statusHelper.js
```diff
@@ -0,0 +1,114 @@
...
+    executorStatus: bool,
+    systemsStatus: {
+        "pending": num,
+        "running": num,
+        "success": num,
+        "failure": num,
+        "canceled": num
+    }
+*/
+export const StatusSummary = ({ executorStatus, permission, needsTooltip, onCancel, counts }) => {
```
jharting, line 66: This function seems to be used for both Runs and Executors (which is good). It needs to account for a fact that in the former case 'acked' is also a possible status, see https://github.com/RedHatInsights/insights-remediations/blob/master/src/api/openapi.yaml#L625 which should be considered "running" when normalizing `runningCount`

src/components/statusHelper.js
```diff
@@ -0,0 +1,114 @@
...
+        </Flex>
+    );
+
+    if (needsTooltip && executorStatus) {
+        return <Tooltip
+            position='right'
+            className='ins-c-status-tooltip'
+            enableFlip
+            content={
+                <span>{ `Run: ${executorStatus.charAt(0).toUpperCase() + executorStatus.slice(1)} |
```
jharting, line 103: ideally this tooltip would list all possible executor statuses when used for a playbook run https://github.com/RedHatInsights/insights-remediations/blob/master/src/api/openapi.yaml#L625 (i.e. including Acked) and all possible system statuses when used for a playbook run executor https://github.com/RedHatInsights/insights-remediations/blob/master/src/api/openapi.yaml#L630 (i.e. without Acked)

src/components/RemediationActivityTable.js
```diff
@@ -0,0 +1,109 @@
...
+                                        <Link to={ `/${remediation.id}/${playbooks.id}/${e.executor_id}` }>
+                                            { e.executor_name }
+                                        </Link>
+                                    },
+                                    e.system_count,
+                                    { title: <StatusSummary
+                                        executorStatus={ normalizeStatus(e.status) }
+                                        counts={ e.counts }
+                                        permission={ permission }
+                                        // onCancel={ () => console.log('cancel') } TODO add back
```
jharting, line 64: Cancelation will only be possible at the playbook_run level (you can cancel the entire run but not one particular executor in a running run)

src/components/ExecutorDetails.js
```diff
@@ -0,0 +1,311 @@
...
+            success: 0,
+            failure: 0,
+            canceled: 0 });
+
+    const renderMain = (status) => ({
+        running: <Main>
+            <Stack gutter="md">
+                <Card>
+                    <CardHeader className='ins-m-card__header-bold'>
+                        <Button
```
jharting, line 145: This looks differently in the mocks: https://marvelapp.com/j3b9b7e/screen/66225551
* should be a button with icon
* can it be on the same "row" as the pagination controls i.e. not on its own row?

src/components/ExecutorDetails.js
```diff
@@ -0,0 +1,311 @@
...
+                            variant='link' onClick={ () => downloadPlaybook(remediation.id) }>
+                            Download Playbook
+                        </Button>
+                    </CardHeader>
+
+                    <CardBody>
+                        { InventoryTable && <InventoryTable
+                            ref={ inventory }
```
jharting, line 153: The mocks have a "filter by name" search box. It's possible to filter hosts by name using the `ansible_host` query param in https://cloud.redhat.com/docs/api/remediations#operations-remediations-getPlaybookRunSystems

src/components/ExecutorDetails.js
```diff
@@ -0,0 +1,311 @@
...
+                            items={ orderBy(systems, [ s => getSystemName(s), s => s.id ]) }
```
jharting, line 154: The server returns the systems order by system name by default. The order can be controlled with the `sort` query param (see https://cloud.redhat.com/docs/api/remediations#operations-remediations-getPlaybookRunSystems)

src/components/ExecutorDetails.js
```diff
@@ -0,0 +1,311 @@
...
+                            onRefresh={ onRefresh }
+                            page={ page }
+                            total={ systems.length }
```
jharting, line 157: this will need to be updated to support real server-side pagination

src/components/ExecutorDetails.js
```diff
@@ -0,0 +1,311 @@
...
+                            onExpandClick={ (_e, _i, isOpen, { id }) => {
+                                getPlaybookRunSystemDetails(remediation.id, run_id, id);
+                                onCollapseInventory(isOpen, id);
+                            } }
+                        /> }
+                    </CardBody>
+                </Card>
+            </Stack>
+        </Main>,
+        failure: <Main>
```
jharting, line 207: even on failure let's show the inventory table so that users have a chance to see what went wrong

src/components/ExecutorDetails.js
```diff
@@ -0,0 +1,311 @@
...
+                            onRefresh={ onRefresh }
+                            page={ page }
+                            total={ systems.length }
+                            perPage={ pageSize }
+                            tableProps={ { onSelect: undefined } }
+                            expandable
+                            onExpandClick={ (_e, _i, isOpen, { id }) => {
+                                if (isOpen) {
+                                    getPlaybookRunSystemDetails(remediation.id, run_id, id);
+                                    setRefreshInterval(setInterval(() => getPlaybookRunSystemDetails(remediation.id, run_id, id), 10000));
```
jharting, line 164: When there is an update the table switches to skeleton and then renders again. Would it be possible just for the log to get updated without the "reload" effect?

src/components/ExecutorDetails.js
```diff
@@ -0,0 +1,311 @@
...
+            children: <Skeleton size='lg' />
+        })));
+    }, [ playbookRunSystems ]);
+
+    useEffect(() => {
+        setSystems(() => ([
+            ...systems.filter(system => system.id !== playbookRunSystemDetails.system_id),
+            { ...systems.find(system => system.id === playbookRunSystemDetails.system_id),
+                isOpen: true,
+                children: <SyntaxHighlighter language="yaml" style={ dark }>
```
jharting, line 126: This is great! When we're in the running state and the console is being updated using the interval would it be possible to add some indication that it is live? (e.g. a small spinner below the console?) I know it's not in the mocks just curious if you had any ideas

src/components/ExecutorDetails.js
```diff
@@ -0,0 +1,311 @@
...
+                        <Link to={ `/${remediation.id}` }> { remediation.name } </Link>
+                    </BreadcrumbItem>
+                    <BreadcrumbItem>
+                        <Link to={ `/${remediation.id}/${run_id}` }>  <DateFormat type='exact' date={ playbookRun.data.created_at } /> </Link>
+                    </BreadcrumbItem>
+                    <BreadcrumbItem isActive> { executor.executor_name } </BreadcrumbItem>
+                </Breadcrumb>
+                <Stack gutter>
+                    <StackItem>
+                        <PageHeaderTitle title={ executor.executor_name } />
```
jharting, line 241: When running there should be the running icon (see https://marvelapp.com/j3b9b7e/screen/66225551)

src/components/ExecutorDetails.js
```diff
@@ -0,0 +1,311 @@
...
+                            perPage={ pageSize }
+                            tableProps={ { onSelect: undefined } }
+                            expandable
+                            onExpandClick={ (_e, _i, isOpen, { id }) => {
+                                if (isOpen) {
+                                    getPlaybookRunSystemDetails(remediation.id, run_id, id);
+                                    setRefreshInterval(setInterval(() => getPlaybookRunSystemDetails(remediation.id, run_id, id), 10000));
+                                }
+                                else {
+                                    clearInterval(refreshInterval);
```
jharting, line 167: this does not seem to get cleared when the user closes the expanded row neither when I navigate to a different part of remediations

jharting:
Screenshot_20200403_110244

The table seems to overflow the page size (notice the upper left corner)

jharting:
@ryelo the auto-reloading of activity tab when a playbook is running is awesome. The only minor thins I found is that one of those activity tab table rows is expanded it retracts on the automatic reload

jharting:
@PanSpagetka I run into a weird race condition from time to time. If I open e.g. a failed executor and then a succeeded one I see this:

Screenshot_20200406_102240
Screenshot_20200406_102308

note that the `dhcp-*` host does not even belong to this executor so it looks like some stale data is kept in redux?

jharting:
Screenshot_20200406_102435

Also the page crashed once with this error ^^

jharting:
@PanSpagetka I can confirm the system/executor status are now properly updated when the playbook run finishes on a system. However I noticed two things:
* when the system is finished the row collapses
* when the system is expanded again it starts to poll the API again (even though it is finished already)

jharting:
@ryelo @PanSpagetka here's a summary of pending problems I am aware of:

* [ ] the system list race condition (@PanSpagetka saw that yesterday during the demo so I assume you know which one)

Screenshot_20200408_101412

* [ ] the filters/buttons are not on the same row as pagination
* [ ] no padding between the filter and the button

* [ ] when a system row of a finished system is expanded it starts to poll (even though the system is not running)

* [ ] there should be a spinner at the end of the console output if the system is running

Screenshot_20200408_101438
Screenshot_20200408_101451

* [ ] when I expand a row it's order in the list changes

* [ ] invalid property warnings - there are quite a few of these. I know these themselves do not cause the page to fail but the fact we set a property in a way that's not valid may indicate that some of the behavior we intended is actually silently not working. Plus these should all be easy to fix:

Screenshot_20200408_103241
Screenshot_20200408_103332
Screenshot_20200408_103349
Screenshot_20200408_103410

jharting:
These are **totally optional** but would make me happy :-)
* [ ] on RemediationDetail can we make the selected tab part of the URL so that I can send somebody a link that, when opened, switches the tab to Activities

PR #148, opened 2020-04-06
Add cancel playbook functionality

ryelo:
This also updates redux which required more work. Plus, sends the new store over to the inventory component, so I'd like @karelhala to also take a look


jharting:
@ryelo are you planning on adding the confirmation modal? https://marvelapp.com/j3b9b7e/screen/63670935

Besides the activity table the cancel link should also go the playbook run detail view (https://marvelapp.com/j3b9b7e/screen/66225551)

src/components/RemediationActivityTable.js
```diff
@@ -2,6 +2,9 @@ import React, { useState, useEffect, useContext } from 'react';
 import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
 
+import { useDispatch } from 'react-redux';
+import * as actions from '../actions';
```
karelhala, line 6: You can import just `cancelPlaybookRuns` and not all actions.

ryelo:
@jharting I didn't even see that version of the mocks. I'll also add that.

PR #155, opened 2020-04-13
Low hanging fruit papercuts (phase 1)

ryelo:
Overall
- Playbook summary sentence case - Should be “Playbook summary” 
- Buttons should be sentence case (“Execute playbook”, “Download playbook”)

Actions
- Labels look like they are 16px, should be 14px heading label 
- Auto reboot to be 2 words
- “Issues” tab should be “Actions” to match what is shown. 
- "Remove System" Should be not red text as per UXD guidelines
- System modal align buttons to the left
- Issues table Remove action should be in a secondary button style
- activity table Spacing in the expand pane is off. 
- status tooltip says “systems” after all the numbers
- fixed floating tooltip - should be closer to the actual content. Hopefully centered on it.
- Status text should have min-width for the label so we can line up all these things (85px)
- adds a banner when receptor is not configured

Executor Details
- Label style issue on details page
- Results by connection not correct title level
- Missing a level of the breadcrumb - “Remediations” then playbook then run
- Executor Detials Header details are actually supposed to be a bit diff (Status is first, followed by “Playbook run by” and “Playbook run on”)
- Toolbar alignment all sorts of off
- Download paybook should be sentence case - “Download playbook”
- Alignment issue on tags

Playbook Log
- Missing the title on this - should have title of “Playbook log” in xl

Execute Playbook modal
- Capitalization of playbook - Should be sentence case

### DIFF
--- a/src/components/ActivityDetails.js
+++ b/src/components/ActivityDetails.js
@@ -13,7 +13,8 @@ import {
     Card, CardHeader, CardBody,
     Stack, StackItem,
     Breadcrumb, BreadcrumbItem,
-    Split, SplitItem
+    Split, SplitItem,
+    Title
 } from '@patternfly/react-core';
 import {
     Table,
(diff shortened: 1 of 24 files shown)


ryelo:
Updated that tooltip 😄

PR #165, opened 2020-04-28
add url to NotConfigured

ryelo:


aleccohan:
@Jason-RH is this the impl that adds a link to the "Enable Cloud Connector" Banner?

Jason-RH:
@aleccohan No this is for an EmptyState that shows up in the Activity tab when receptor is not configured. We still need to go through and audit all of our links and make sure we've got the right ones in the right spot, but currently this state has a link button with no link.

Jason-RH:
@aleccohan Yeah adding in a "component" property will fix that. I can update it

PR #168, opened 2020-04-29
Fix js error at System for Action modal

PanSpagetka:
Fix js error when loading Remediation page.
https://projects.engineering.redhat.com/browse/RHCLOUD-5312


src/components/SystemForActionButton.js
```diff
@@ -60,8 +60,11 @@ const SystemForActionButton = ({ issue, remediation, onDelete }) => {
     };
 
     useEffect(() => {
-        loadInventory();
-    }, []);
+        if (open) {
```
karelhala, line 63: I don't think this is the proper way, this will eventually create new instance of inventory everytime user openes/closes modal.

PR #172, opened 2020-05-28
Add line breaks to playbook run status tooltip

PanSpagetka:
Add line breaks to playbook run status tooltip. Group number by status.

@katierik Does this look good to you?

https://projects.engineering.redhat.com/browse/RHCLOUD-6695

### Before:
Screenshot from 2020-05-28 09-59-06

### After:
Screenshot from 2020-06-02 09-55-00


ryelo:
I don't know if we want it to say:
"Failed, Canceled: 1 system"

Because failed and canceled can actually be two different events that I think are worked on being separated. I think if a value is 0, either hide it entirely or show a 0 next to it.

katierik:
Re: failed/canceled, I believe the red (x) on the UI is representing both (hence the combo). 

They're being separated on the backend or what is happening there?

ryelo:
Yeah they both shared the red X, but in line with Alec's PR here: https://github.com/RedHatInsights/insights-remediations-frontend/pull/173, it seems like canceled should be called out separately.

Failed and Canceled are two different values in this case (although we group them together as "failure").

So, a user could have something like:
Failed: 1, Canceled: 1 -- which would result in a overall "Failure" of 2 systems.

PanSpagetka:
I am not sure if I have understood it correctly, is this what you have meant?
Screenshot from 2020-06-02 09-55-00

PR #178, opened 2020-06-18
Correcting remediation type when action is from advisor.

Jason-RH:
Also correcting the url formation for advisor actions.


Jason-RH:
@aleccohan Yep!

ryelo:
Travis has it passing, so if the result isn't reported here, i'll just admin merge

https://travis-ci.org/github/RedHatInsights/insights-remediations-frontend/builds/699778311

PR #195, opened 2020-08-25
RHCLOUD-1127 updating empty states for remediations and actions

Jason-RH:
Jira: https://projects.engineering.redhat.com/browse/RHCLOUD-1127
Updated UX review doc: https://docs.google.com/spreadsheets/d/10HLclDalIEXbiQ2cnFeMjwIl4ffge7F716ptrnBW4N8/edit#gid=0


Jason-RH:
@aleccohan "This playbook is empty" when no actions are in the playbook, and "no actions" when the playbook has actions but they're not shown due to an active filter. This is how it currently is in remediations so I didn't change that. The "Updated Design" is the empty state content. The tabs do look different because of a different toolbar/table, but that's unrelated to this PR.

PR #199, opened 2020-09-02
Add important to progress display property.

Hyperkid123:
task: https://projects.engineering.redhat.com/browse/RHCLOUD-8832

We are dealing with the CSS loading issue. The styles break after you open the inventory component because it adds a new style tag to the DOM head. CSS always gives priority to the last styles sheet with particular CSS rule (in this case its `.pf-c-progress` display property). The display property defined in `PlaybookCard.scss` is overridden by the default progress display property to `display: grid` which is appended by the lazy-loaded inventory component somewhere at the end of the head tag.

Unfortunately, we can't fix it inside the inventory component, because we would have to remove the style import, which would break the component styling in apps which do not add the progress styling or if its using code splitting and the browser did not load any component with the progress component in it.

Adding `!important` breaks the default rule behavior and will give priority to the last rule with `!important` flag.

@Jason-RH @ryelo @aleccohan have you noticed any other styling issues after the inventory component is loaded?

important-card

### DIFF
--- a/src/components/PlaybookCard.scss
+++ b/src/components/PlaybookCard.scss
@@ -11,6 +11,7 @@
             margin-right: var(--pf-global--spacer--sm);
             position: relative;
             bottom: 2px; // because we want to have this inline with the text, we can't use flex
+            font-weight: var(--pf-global--FontWeight--normal) !important;
         }
         &--name { 
             word-break: break-word;
@@ -33,7 +34,7 @@
     }
 
     &__progress {
-        display: block;
+        display: block !important;
         --pf-c-progress__bar--before--BackgroundColor: var(--pf-global--BackgroundColor--dark-100);
         --pf-c-progress__indicator--BackgroundColor: var(--pf-global--primary-color--100);
         &--success {


ryelo:
I think the `Archived` label also gets bolded after inventory is opened

Hyperkid123:
> I think the `Archived` label also gets bolded after inventory is opened

Yep, its gets bolded, but it is not a label but a badge, and the badge component should be bolded. This is caused by an outdated PF version. There are two ways this should be eventually fixed. Either update the PF version or switch the component to Label.

For now, I will just override the CSS so it stays the same. But the PF version should be probably updated.

Jason-RH:
@Hyperkid123 I'm actually no longer seeing the styles break in beta environments. Do you mind taking a look and see if you're still seeing it? I'm not sure if something changed elsewhere.

Hyperkid123:
@Jason-RH there were some updates to the chrome build process. It now uses almost the same process as APPS. Let me double-check.

Hyperkid123:
:tada: It seems to be working. I don't see additional style tags on ci.beta env.

Hyperkid123:
The issue has magically disappeared (not really). So I guess I can close this one.

Jason-RH:
Thanks for looking into this! Magically disappearing issues are the best kind.

PR #202, opened 2020-09-17
Fixes/card pagination

Jason-RH:
This PR adds the new 'hide_archived' query parameter to loadRemediations. This enables proper pagination values when hiding the archived playbooks

### DIFF
--- a/src/routes/Home.js
+++ b/src/routes/Home.js
@@ -61,7 +61,12 @@ function Home () {
 
     function load () {
         const column = SORTING_ITERATEES[sorter.sortBy];
-        loadRemediations(column, sorter.sortDir, filter.value, pagination.pageSize, pagination.offset, showArchived);
+        if (showArchived) {
+            loadRemediations(column, sorter.sortDir, filter.value, pagination.pageSize, pagination.offset);
+        } else {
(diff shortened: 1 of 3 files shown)


src/routes/Home.js
```diff
@@ -70,7 +75,7 @@ function Home () {
         if (remediations.status === 'fulfilled' && filter.value === filterText) {
             setShouldUpdateGrid(true);
         }
-    }, [ sorter.sortBy, sorter.sortDir, filter.value, pagination.pageSize, pagination.pageDebounced ]);
+    }, [ sorter.sortBy, sorter.sortDir, filter.value, pagination.pageSize, pagination.pageDebounced, showArchived ]);
```
aleccohan, line 78: @Jason-RH is this **showArchived** supposed to be a hideArchived?

Jason-RH: Nope. A little confusing since we switched to using hideArchived on the backend, but showArchived is what gets changed when toggling the archive view.

aleccohan: Ah!  I see, thank you for the clarification!

aleccohan (approved):
Looks great to me!

PR #207, opened 2020-10-27
Get resolved issue count from filtered issues

Jason-RH:
This change allows the frontend (specifically within the chart on the details view) to calculate the number of resolved issues from a filtered list of issues. This is primarily to ensure that the number of resolved issues does not exceed the number of issues shown in the table below the chart (this could occur if you have an issue that has been resolved, but is no longer found in advisor/vulnerability/etc). Further work/design will be needed for these edge cases.


aleccohan (approved):
Looks good to me!  Counts are matching up great!

PR #211, opened 2020-12-08
Webpack upgrade

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-11107

This is in preparation for chrome 2.0 migration. After this update, the UI will be compatible with chrome 2.0

webpack upgrade PR: https://github.com/RedHatInsights/insights-remediations-frontend/pull/211/commits/d064906090dc659b8dc6264d41e14640fbb593e8

rest is linting


karelhala (commented):
Remediations is using inventory, I don't see it updated, do you want me to do a followup PR to do it, or will you do it?

Hyperkid123:
@karelhala I will update it. Is there anything special I should know about?

Hyperkid123:
@karelhala inventory was updated. I would prefer if somebody went through again and check if the inventory components are working as expected. I believe this is the first application using the new system.

karelhala (approved):

PR #212, opened 2020-12-10
Add more OUIA stuff

ryelo:


src/Routes.js
```diff
@@ -6,18 +6,20 @@ import RemediationDetails from './routes/RemediationDetails';
 import ActivityDetails from './components/ActivityDetails';
 import ExecutorDetails from './components/ExecutorDetails';
 
-const InsightsRoute = ({ component: Component, rootClass, ...rest }) => {
+const InsightsRoute = ({ component: Component, rootClass, pageType, ...rest }) => {
     const root = document.getElementById('root');
     root.removeAttribute('class');
     root.classList.add(`page__${rootClass}`, 'pf-l-page__main');
     root.classList.add(`page__${rootClass}`, 'pf-c-page__main');
+    root.setAttribute('data-ouia-page-type', pageType);
```
Hyperkid123, line 14: We should not be setting these attributes directly. The whole `rootClass` stuff is deprecated in chrome 2 apps because it can mess up other applications.

You can use `insights.chrome.appAction('foo')` instead on component mount.

Hyperkid123 (changes requested):

PR #224, opened 2021-04-21
Add systems tab to remediations detail

karelhala:
JIRA: https://issues.redhat.com/browse/RHCLOUD-13786

Screenshot from 2021-04-21 15-58-43
Screenshot from 2021-04-21 15-58-53
Screenshot from 2021-04-21 15-59-01

### DIFF
--- a/src/store/reducers.js
+++ b/src/store/reducers.js
@@ -4,6 +4,7 @@ import { ACTION_TYPES } from '../constants';
 import { applyReducerHash } from '@redhat-cloud-services/frontend-components-utilities/files/ReducerRegistry';
 import flatMap from 'lodash/flatMap';
 import uniq from 'lodash/uniq';
+import { RebootColumn, IssuesColumn } from '../components/SystemsTable';
 
 function issuesToSystemsIds(issues) {
   return uniq(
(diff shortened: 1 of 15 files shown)


src/store/reducers.js
```diff
@@ -26,6 +27,79 @@ function computeRebootStats(remediation) {
...
+                displayName={display_name}
+              />
+            ),
+            props: { width: 15 },
+          },
+          {
+            key: 'rebootRequired',
+            title: 'Reboot required',
+            // eslint-disable-next-line react/display-name
+            renderFunc: (rebootRequired) => (
```
Hyperkid123, line 87: Can we not store react components in redux?

src/store/reducers.js
```diff
@@ -26,6 +27,79 @@ function computeRebootStats(remediation) {
   };
 }
 
+export const remediationSystems = ({ LOAD_ENTITIES_FULFILLED }) =>
+  applyReducerHash({
+    [ACTION_TYPES.SELECT_ENTITY]: (state, { payload }) => {
+      const selected = state.selected || new Map();
```
Hyperkid123, line 33: Maps are mutable and can cause issues within redux. Can we use a plain object instead?

src/store/reducers.js
```diff
@@ -26,6 +27,79 @@ function computeRebootStats(remediation) {
...
+        } else if (payload.id === -1) {
+          selected.clear();
+        } else {
+          selected.delete(payload.id);
+        }
+      }
+
+      return {
+        ...state,
+        selected: new Map(selected),
```
Hyperkid123, line 55: ^^

src/store/reducers.js
karelhala: The map is specifically used so we don't have to run `filters` and `finds` and removing the columns. I'll mark this as `TODO: refactor` and we can fix it later on.

src/store/reducers.js
karelhala: I'll try looking at `cellTransformer` so we don't have to use the `renderFunc`

karelhala: Mhh `cellFormatters` wouldn't help us either as that's the same as `renderFunc` and inventory expects columns to be defined in redux and not trough props.

Hyperkid123 (approved):

PR #236, opened 2021-05-18
Show id as fallback in system table

PanSpagetka:
Some systems doesn't have defined `fqdn` use `name` or `id` as fallback.


aleccohan (approved):
LGTM!

PR #242, opened 2021-05-25
Migrate to npm7

Hyperkid123:
@fhlavac @PanSpagetka can you please test it locally and check that stuff is still working? I did run it locally but I am not that familiar with the whole app as you are :slightly_smiling_face:


fhlavac (approved):
It looks good to me! Thank you :-)

PR #250, opened 2021-06-16
change on status display rendering

arivepr:
This looks to close: [RHCLOUD-10604](https://issues.redhat.com/browse/RHCLOUD-10604)

Before: 
On dashboard, an action would look like this if not completely done: 
Screen Shot 2021-06-16 at 9 00 29 AM

While on the playbooks side on Remediations, it looks like:
Screen Shot 2021-06-16 at 9 02 13 AM

-------

This proposed fix aims to avoid applying the completed status to those that haven't even ran, or have nothing to run, thus keeping things more in line with what appears on dashboard: 

Screen Shot 2021-06-16 at 9 07 28 AM

------

**QUESTION**
This is the proposed fix for the Remediations playbook side of things, any suggestions as far as Dashboard goes? Been looking through it, but don't know if it's worth changing after these changes.


Hyperkid123:
@arivepr what is there such a huge lock file change? I don't see anything in the `package.json` changed. What node and npm version are you using?

arivepr:
> @arivepr what is there such a huge lock file change? I don't see anything in the `package.json` changed. What node and npm version are you using?

Holy, I didn't notice that initially. I'm currently on npm 6.14.8, node  v14.15.1

ryelo:
@arivepr You should just be able to remove that file from the commit/reset that file to what it is on master. In the future, you can always just not commit the lockfile if you know you didn't change anything in the package.json.

Hyperkid123:
@arivepr ah you should be at node 15+ and npm7+. It's strange because the package.json has versions defined and should throw an error when you try using different versions... If you don't have multiple versions of node installed I suggest using `nvm` to manage that.

But as @ryelo  said, just remove the lock file from the commit and we are good it think,

Hyperkid123 (approved):

arivepr:
Thanks for letting me know guys! I do have nvm, so I'll just set that up moving forward.

PR #252, opened 2021-06-30
Adding empty state to execute modal without systems

arivepr:
This PR aims to close [this card](https://issues.redhat.com/browse/RHCLOUD-14910)

Screen Shot 2021-06-30 at 11 03 08 AM

_**Green square around x button is from trying to take screenshot**_

Also, any feedback to make the conditional rendering of the table portion is welcomed with open arms. Couldn't figure out a way to make it simpler/cleaner, and did not want to spin wheels on it :D

### DIFF
--- a/src/components/Modals/ExecuteModal.js
+++ b/src/components/Modals/ExecuteModal.js
@@ -13,7 +13,13 @@ import {
   ExpandableSection,
   List,
   ListItem,
+  Title,
+  EmptyState,
+  EmptyStateBody,
+  EmptyStateIcon,
+  EmptyStateSecondaryActions
 } from '@patternfly/react-core';
+import DesktopIcon from '@patternfly/react-icons/dist/js/icons/desktop-icon';
 import { downloadPlaybook } from '../../api';
 import { styledConnectionStatus } from '../statusHelper';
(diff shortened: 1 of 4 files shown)


src/components/Modals/ExecuteModal.js
```diff
@@ -149,42 +150,57 @@ export const ExecuteModal = ({
...
+              >
+                Download playbook
+              </Button>,
+              isDebug() ? (
+                <Button
+                  key="reset-etag"
+                  onClick={() => setEtag('test')}
+                  ouiaId="reset-etag"
+                >
+                  Reset etag
```
Hyperkid123, line 190: typo in `etag`?

Hyperkid123 (commented):
Just found a little typo? Or is it supposed to be like that?

src/components/Modals/ExecuteModal.js
arivepr: not sure what this is, I don't think I wrote that directly

ryelo: etag is correct

Hyperkid123 (approved):

PR #256, opened 2021-07-21
Add modules to control Button and wizard

karelhala:
### Move remediations wizard over

Since we want to make the transition to a new team as easy as possible, we have to move the remediations button and wizard to remediations application. This is first PR of a series, firstly we'll promote this code to all envs. After that we'll change the way how remediations is loaded in FEC using async component and after that we can also change the loader in chrome to fully cover any corner case.

This PR is just copy + pase of original code with a few tweaks of how data are loaded, instead of firing `openWizard` we'll just wrap the wizard in condition and if no data are provided don't render the wizard. Once user clicks on remediate we'll collect the data the same way as previously and launch the wizard.


src/modules/RemediationsModal/RemediationsWizard.js
```diff
@@ -0,0 +1,281 @@
...
+      Object.values(value).filter((value) => typeof value !== undefined).length
+        ? undefined
+        : 'At least one system must be selected. Actions must be associated to a system to be added to a playbook.',
+  };
+
+  return (
+    <Fragment>
+      {state.schema && !state.submitted ? (
+        <FormRenderer
+          className="remediationsss"
```
fhlavac, line 144: This is probably extra

src/modules/RemediationsModal/RemediationsWizard.js
```diff
@@ -0,0 +1,281 @@
...
+            [ISSUES_MULTIPLE]: [],
+            [SYSTEMS]: {},
+            [MANUAL_RESOLUTION]: true,
+            [SELECTED_RESOLUTIONS]: {},
+            [EXISTING_PLAYBOOK_SELECTED]: false,
+          }}
+          componentMapper={{
+            [componentTypes.WIZARD]: {
+              component: WizardMapper,
+              className: 'remediations',
```
fhlavac, line 161: Is this class needed?

fhlavac (approved):
Just one nitpick. Went through the code and tested it locally. Looks good to me! thanks for the code cleaning :-)

PR #274, opened 2021-10-21
Hotfix review step by conditional inv rendering

karelhala:
### Fix DDoS error

When navigating to review step with many systems the DDoS protection kicks in because we are rendering inventory for each collapsed row. This is not good because each inventory table performs 2 requests for chrome's fed modules. This PR fixes such issue by rendering inventory table only if row is expanded.


Hyperkid123 (approved):
Tested it in browsers and it works. Inventory starts to load only after a row is expanded. We will have to promote these changes to prod.

PR #283, opened 2022-01-27
Substitute * in RBAC permission for Remediate button

leSamo:
Remediations here were matching permissions literally and not substituting *, which means for example when user has `remediations:*:*` permission it won't let them do the Remediate action requiring `remediations:remediation:write`, even though it should, solution is to match any substring between : to *, instead of doing whole string matching with ===

This lead to Remediation button not working when it should in Vulnerability app.


mkholjuraev (approved):
LGTM. Nice and clean. I like it.

PR #291, opened 2022-02-16
Display no data modal when no remediable issues are passed

leSamo:
Ticket: https://issues.redhat.com/browse/VULN-2041
Mockup: https://marvelapp.com/prototype/gbhbja0/screen/83265660

### Possible situations
1. User has no items selected -> Remediate button should be disabled - this is controlled by apps
2. User has selected items, but none are remediable (remediation is manual or there is none) -> Remediation button
3. User has selected items and some or all are remediable -> show Wizard and remediate normally

This PR implements situation 2.

Apps using Remediation button component are responsible for disabling/enabling the button. Apps should send `false` instead of data object, when nothing selected is remediable.

## Mockup:
mockup

## Actual:
actual

### DIFF
--- a/src/modules/RemediationsButton.js
+++ b/src/modules/RemediationsButton.js
@@ -6,6 +6,7 @@ import validate from './RemediationsModal/validate';
 import { CAN_REMEDIATE, matchPermissions } from '../Utilities/utils';
 import { Button, Tooltip } from '@patternfly/react-core';
 import RemediationWizard from './RemediationsModal';
+import NoDataModal from './RemediationsModal/NoDataModal';
 
 const RemediationButton = ({
   isDisabled,
@@ -16,6 +17,7 @@ const RemediationButton = ({
 }) => {
   const [hasPermissions, setHasPermissions] = useState(false);
   const [remediationsData, setRemediationsData] = useState();
+  const [isNoDataModalOpen, setNoDataModalOpen] = useState(false);
 
   useEffect(() => {
     insights.chrome.getUserPermissions('remediations').then((permissions) => {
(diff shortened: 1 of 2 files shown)


johnsonm325 (commented):
This looks good to me. Is the expectation that other apps can choose to disable the button even if they have selected manual remediation? If you want to ensure that other apps aren't passing `isDisabled=true` even when manual remediation is selection, then I might suggest this for line 47:

`isDisabled={ !data ? false : isDisabled }`
that may be able to be expressed simpler. But basically we want to say, if `data === false` then don't disable the button. If data is empty object or has content, then use `isDisabled`.

johnsonm325:
> @johnsonm325 thank you for the review, all apps should behave consistently so it might sense to restrict the `isDisabled` prop; but on the other hand adding this condition to `isDisabled` might confuse future developers (why doesn't changing `isDisabled` prop do anything?), so my proposal would be to remove `isDisabled` prop and enable/disable button based on received data:
> 
> * `data === null` - no items selected -> disable button
> * `data === false` - items selected, but none are remediable -> enable button, on click show no data modal
> * `data.issues.length > 0` - items selected and some or all are remediable -> enable button, on click show remediation wizard
> 
> There is no need to have `isDisabled` prop with this setup, so we'd remove its usage from apps, but we could leave this prop for future use and make it overwrite the conditions above. Sounds sensible?

Absolutely. I think that's a great idea.

MichalSajdik (approved):

leSamo:
I forgot data provider is actually a promise that's only called upon clicking the button, so I guess calling data provider each time to decide whether to enable/disable button is a bad idea.

PR #293, opened 2022-02-23
fix: Updated Playbook Card design when archived

AsToNlele:
ESSNTL-786

Archived Remediations now look as requested:
image

### DIFF
--- a/src/components/PlaybookCard.js
+++ b/src/components/PlaybookCard.js
@@ -1,7 +1,6 @@
 import React, { useState } from 'react';
 import { useDispatch } from 'react-redux';
 import {
-  Badge,
   Card,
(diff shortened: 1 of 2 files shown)


leSamo (changes requested):
Thank you for providing pictures, it helps a lot;

Progress bar looks correct, I noticed there are two discrepancies with mockup, the archieved label style is compact, but it should not be (Label Patternfly element, controlled by `isCompact` prop) and the playbook name should have right margin so there is a space between playbook name and archieved label

AsToNlele:
image

Sorry! Should be fine now.

leSamo (approved):
Perfect

PR #304, opened 2022-03-16
fix: Update reference to remediations object

johnsonm325:
This bug happens in beta when you visit the remediations page. The console specifies that n.reduce is not a function. This is because the api returns `remediations: { value: { data: { data: [] } } }` instead of `remediations: { value: { data: [] } }`.


johnsonm325:
Not sure why the package-lock has updated. I'm using node version 15. If this needs to be fixed, let me know and I'll update.

bastilian:
Can we find out why the structure changed and there is now another level of `data` and avoid working around a potential bug somewhere else and introduce a future bug?

Re the package-lock.json: Let's do a bump to node 16 separatly. For now, make sure to use npm version 6 when installing packages for remediation.

johnsonm325:
@bastilian It appears that the build process is using npm version 7

PR #312, opened 2022-04-28
fix: ESSNTL-2670 - fix playbook delete button

AsToNlele:
[ESSNTL-2670](https://issues.redhat.com/browse/ESSNTL-2670)
Previously when you cancelled a deletion of a playbook you would get a notification of deletion.
Fixed dialog text

Before:
image

After:
image


mkholjuraev (approved):
LGTM, tested locally! thanks for the PR.

PR #316, opened 2022-06-16
fix: ESSNTL-2827 - Playbook output has a new line for each line

AsToNlele:
[ESSNTL-2827](https://issues.redhat.com/browse/ESSNTL-2827)

Playbook output has a new line for each line

I'm not sure if this is the best way to handle this, open for suggestions

If you want to review this PR I can provide you an account with this playbook

Before:
image

After:
image

### DIFF
--- a/src/components/RemediationsLogViewer.js
+++ b/src/components/RemediationsLogViewer.js
@@ -4,6 +4,9 @@ import { LogViewer, LogViewerSearch } from '@patternfly/react-log-viewer';
 import { Toolbar, ToolbarContent, ToolbarItem } from '@patternfly/react-core';
 
 const RemediationsLogViewer = ({ data }) => {
+  // New line after each line
+  data = data.replaceAll('\n', '\n\n').slice(0, -1);
+
   return (
     <LogViewer
       data={data}


Ichimonji10:
This is what ansible-playbook's output would look like if executing in an actual console:
```
[WARNING]: provided hosts list is empty, only localhost is available. Note that
the implicit localhost does not match 'all'

PLAY [Set owner and permissions on /etc/sshd/sshd_config to root:root 0600] ****

TASK [Gathering Facts] *********************************************************
ok: [localhost]

TASK [Set the owner and permissions of ssh config file to root:root 0600] ******
changed: [localhost]

PLAY RECAP *********************************************************************
localhost                  : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

Ideally, we should match that. It's possible that the web UI is correctly formatting what it's being given, and whatever is submitting these logs is mangling ansible-playbook's outputs. However, I also wonder if there's some encoding issue - perhaps one app uses `\r\n` to denote newlines, and some other app uses `\n`?

rexwhite:
I looked at what the api is returning to the UI for console output and it looks like most of the formatting problems are upstream of the UI (i.e. it's not the UI's fault).  The UI does seem to be wrapping long lines, which might not be the best idea, but I think this particular PR only makes the formatting worse.  I'll investigate the source of the upstream formatting issues and open a new UI jira if we want to change the log line wrapping behavior.

I'm closing this PR and assigning the ESSNTL-2827 jira to myself to address the upstream issues.

PR #317, opened 2022-06-17
Updating FEC and PF packages, including FEC Wizard removal

gabipodolnikova:


src/components/ResolutionEditButton.js
```diff
@@ -1,59 +1,63 @@
...
-    };
-  }
+const ResolutionEditButton = ({
+  remediation,
+  issue,
+  onResolutionSelected,
+  getResolutions,
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const openModal = () => setIsOpen(true);
```
karelhala, line 14: I don't think we need to have seperate constant for this, we can directly call `setIsOpen(true)` on line 29.

karelhala (approved):

PR #319, opened 2022-07-07
fix ExecutModal connection status

mkholjuraev:
This PR [resolves](https://issues.redhat.com/browse/ESSNTL-1482) the connection status column on ExecuteModal.

It needs to be checked against side effects as I do not have full understanding of connection status use-cases. BUT, I have indentified that RHC hosts do not have availability info and this is breaking `combineStatuses` function result. Thus, I have introduced a single condition to prevent availability checking for RHC hosts only.

These hosts can be used to reproduce the issue, one of them is satellite and another is RHC:

1. ci-vm-10-0-147-6.hosted.upshift.rdu2.redhat.com 
2. ci-vm-10-0-152-181.hosted.upshift.rdu2.redhat.com

Also thre are three remediations for those hosts. To view them, log in to console.redhat.com as the insights-qa user, and search for the following remediations:

jaudet-muslimjon-sat
jaudet-muslimjon-rhc
jaudet-muslimjon-both
The names are indicative of which host(s) the remediations target.

### DIFF
--- a/src/components/ExecuteButton.js
+++ b/src/components/ExecuteButton.js
@@ -19,8 +19,6 @@ const ExecuteButton = ({
   runRemediation,
   etag,
   remediationStatus,
-  getEndpoint,
-  sources,
   setEtag,
   setActiveAlert,
 }) => {
@@ -75,8 +73,6 @@ const ExecuteButton = ({
           issueCount={issueCount}
           runRemediation={runRemediation}
           setEtag={setEtag}
-          getEndpoint={getEndpoint}
-          sources={sources}
           setActiveAlert={setActiveAlert}
         />
       )}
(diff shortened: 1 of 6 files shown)


rexwhite (commented):
This looks like it will work but it also seems like the code is still querying sources directly for the satellite case.  It's no longer necessary for the UI to query sources at all - you can just use the status in the data returned by  `https://console.redhat.com/api/remediations/v1/remediations/3a27368b-80b6-45c4-9a15-c90b9c301244/connection_status`

Currently there are three distinct connection types returned by the `connection_status` api endpoint that will show up in the "results by connection" table:

-   satellite
-   RHC
-   RHC-satellite

The status for each of these types should come directly from the data returned by the `connection_status` api endpoint.

mkholjuraev:
@rexwhite thank you for having a look. Indeed it looks like that request can be safely removed. I believe that availability status also needs to be removed together. In the new commit I have done neccessary clean up. Please have a look and let me know if anything needs to be changed or does not make sense. Thanks

rexwhite (approved):
I'm not super familiar with this code but this looks ok to me.  I'm fine with merging this and promoting it to stage to test.

adonispuente (approved):
PR works as intended, LGTM!

mkholjuraev:
@rexwhite @adonispuente thank you for review!

PR #323, opened 2022-08-18
fix(REMEDY-56): enable to deselect selected remediations

mkholjuraev:
This fixes: https://issues.redhat.com/browse/REMEDY-56.

Users should be able to:

1. Select page
2. Deselect all selected remediations

in remediations table. Selecting all existing remediations will be done later, when there is an API to fetch all IDs of those remediations.


johnsonm325 (approved):
LGTM!

PR #324, opened 2022-09-27
fix(Progress): REMEDY-83 - Fix beta wizard links

tompsota:
resolves https://issues.redhat.com/browse/REMEDY-83

Actually, redirect to stable happened only after second remediation. I suspect that `href` to playbook caused `document.baseURI` to change from `https://stage.foo.redhat.com:1337/beta/` to `https://stage.foo.redhat.com:1337/`. Therefore we have to check and remove `beta/` from `document.baseURI`

### DIFF
--- a/src/Utilities/utils.js
+++ b/src/Utilities/utils.js
@@ -37,8 +37,11 @@ export const getEnvUrl = () => {
   return pathName[1] === 'beta' ? 'beta/' : '';
 };
 
+export const getBaseUri = () =>
+  `${document.baseURI.replace('beta/', '')}${getEnvUrl()}`;
+
 export const remediationUrl = (id) =>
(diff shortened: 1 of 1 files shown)


bastilian:
@tompsota Have you tried to use the `Link` component from `react-router-dom` for the link/button[0] to open the playbook? This should eliminate the need to determine the environment, beta or stable, based on the url.

[0] https://github.com/RedHatInsights/insights-remediations-frontend/blob/master/src/modules/RemediationsModal/steps/progress.js#L155

tompsota:
@bastilian I tried, but that didn't solve the problem

bastilian (approved):
LGTM! Thank you @tompsota!

PR #327, opened 2022-09-28
[REMEDY-219] Fix remediations loading errors

johnsonm325:
To reproduce:
- Navigate to `/insights`
- Click on Remediations in the side nav (may have to do this multiple times)

This issue is intermittent. If the page loads properly, then you'll need to navigate back to `/insights` and reload the page and start the process all over again. Eventually, you'll get the errors.

You will also see this issue of you navigate to `/insights` and then click the "All apps and services" button in the header and choose "Remediations" from the "Ansible Automation Platform" section.

This PR fixes that issue.

### DIFF
--- a/src/AppEntry.js
+++ b/src/AppEntry.js
@@ -4,19 +4,15 @@ import { BrowserRouter as Router } from 'react-router-dom';
 import { Provider } from 'react-redux';
 import { init } from './store';
 import App from './App';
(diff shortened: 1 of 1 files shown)


bastilian (approved):
LGTM! Thank you @johnsonm325!

leSamo (approved):

PR #339, opened 2022-11-15
fix(SystemsTable): REMEDY-265 - Don't set columns via redux store

bastilian:
When opening the Remediations app, and navigate to the "Systems" tab the loading of the Inventory Table will fail without this fix, as it tried to use the columns in the Inventory redux store that have been removed.


mkholjuraev (approved):
LGTM! tested on stage-stable. Systems tab loads for all, no other issue that I noticed.

PR #340, opened 2022-11-15
[REMEDY-219] Fix new loading issue

johnsonm325:
To reproduce the bug:
- Navigate to /insights
- Click Toolkit -> Remediations in the side nave
- After the page loads, click Remediations in the side nave again

It is an intermittent error, and sometimes I have to try it 6 or 7 times to get it to error, but it does eventually do it. This should be fixed.


gkarat:
LGTM, run locally: I could not reproduce this bug anymore with this pull request.

gkarat (approved):

PR #345, opened 2022-12-05
fix(REMEDY-277): extract columns from reducer hash

mkholjuraev:
# Description

Associated Jira ticket: [#REMEDY-277](https://issues.redhat.com/browse/REMEDY-277)

This extracts adding columns from reducers to resolve the error. Instead, columns will be provided by columns callback function.

# How to test the PR

1. Open a remediation detail page by clicking one of the remediation name link
2. visit the 'Activity' tab
3. Open the executer detail page by selecting connection name link
4. Observe that page loads as expected

# Before the change

# After the change

# Dependent work link

# Checklist:


src/components/ExecutorDetails/Columns.js
```diff
@@ -0,0 +1,25 @@
...
+const urlBuilder = inventoryUrlBuilder({ id: 'default' });
+
+export default [
+  {
+    key: 'display_name',
+    title: 'Name',
+    // eslint-disable-next-line
+    renderFunc: (name, id, { fqdn }) => <div><a href={urlBuilder(id)}>{fqdn || name || id}</a></div>
+  },
+  'tags',
```
gkarat, line 14: This value type (string) is somewhat different from what we had before (`state.columns.find((col) => col.key === 'tags')`, object?). I am not quite familiar with the app, and can't say it is bad or good, just suspicious.

gkarat (approved):
Looks OK. Nothing problematic, works locally. However, I left a couple of questions, but they are mostly of curiosity.

src/components/ExecutorDetails/Columns.js
mkholjuraev: This is to consume the tags column from the Inventory app, rather than creating a custom. `mergeColumns` function is designed to handle this.

src/components/ExecutorDetails/Columns.js
```diff
@@ -0,0 +1,25 @@
...
+  {
+    key: 'status',
+    title: 'Status',
+    // eslint-disable-next-line
+    renderFunc: (status) => (
+      <div className="rem-c-status-bar">
```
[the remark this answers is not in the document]
mkholjuraev: To apply the CSS rule from [here](https://github.com/RedHatInsights/insights-remediations-frontend/blob/22d2275e90af4f4168ce84ec0d489b707f140daf/src/components/RemediationActivityTable.scss#L6)

PR #350, opened 2023-01-03
fix(InventoryColumns): use one custom columns array to merge with inventory default columns

mkholjuraev:
# Description

Associated Jira ticket: # (issue)

This resolves the regression on the Executor details page. The regression was caused while resolving REMEDY-277. It appears that Executor details page also uses the mergedColumns function with the obsolete column merging logic. This PR updates the usage of the mergedColumn function to be compatible for both pages. The regression is only on stage env.

# How to test the PR
1. Run the PR against prod env. The dev-webpack.config is prepared for running the PR agains PROD, once approved I will remove this.
2. search for playbook 'mkholjur-test-search'
3. click on the activities tab
4. open one activity detail
5. Observe that the inventory module does not fail to load.

# Before the change

# After the change

# Dependent work link

# Checklist:

### DIFF
--- a/config/dev.webpack.config.js
+++ b/config/dev.webpack.config.js
@@ -29,7 +29,7 @@ const insightsProxy = {
 const webpackProxy = {
   deployment: process.env.BETA ? 'beta/apps' : 'apps',
   useProxy: true,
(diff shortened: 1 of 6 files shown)


Fewwy (approved):
Tested it locally. LGTM

PR #365, opened 2023-04-12
feat(ESSNTL-4496): Replace SmartManager w/ satellite

adonispuente:
# Description
Associated Jira ticket: # ([ESSNTL-4496)](https://issues.redhat.com/browse/ESSNTL-4496)
This replaces all references of Red Hat Smart Manager with Satellite. I've left permissions as is for the time being.

# How to test the PR
Just skim the code and UI for any references to smart manager and verify there are none the user can see.

# Checklist:

- [ x] PR has a short description
- [x ] Tests for the changes have been added


src/components/EmptyStates/ActivityTabUpsell.js
```diff
@@ -20,14 +20,14 @@ const ActivityTabUpsell = () => (
         Get more with Find it Fix it capabilities
       </Title>
       <EmptyStateBody>
-        Upgrade to Red Hat Smart Management to remediate all your systems,
-        across regions and geographies directly from Red Hat Insights.
+        Upgrade to Satellite to remediate all your systems, across regions and
```
leSamo, line 23: Shouldn't Satellite be prepended with "Red Hat" as it was the case with "Red Hat Smart Management" previously?

adonispuente: Good catch, that does sound right

adonispuente:
Not sure what this error means in ci.ext 09:46:00 ERROR: Step ?Publish JUnit test result report? failed: No test report files were found. Configuration error?

Travis is saying the linter is incorrect but locally its passing.

adonispuente:
@leSamo bringing this up again

src/App.js
```diff
@@ -98,7 +98,7 @@ const App = (props) => {
           execute: executePermission,
         },
         isReceptorConfigured,
-        hasSmartManagement,
+        hasSatellite,
```
gkarat, line 101: I would prefer avoiding changing internal variables since this change will, for example, break components here https://github.com/RedHatInsights/insights-remediations-frontend/blob/9dfb812ffcb1088d8201ea1be50063cbea35e2f6/src/components/PlaybookCard.js#L80 and other. Given that this is a context variable, you have to make sure all consumer components talk to the correct variable. However, we can ignore it and only change the client facing texts :-)

gkarat (approved):

gkarat (changes requested):
oops, wrong button :D

src/App.js
mkholjuraev: Smart management gates are removed. RHIF-245, we are safe here as of yesterday.

mkholjuraev (approved):
LGTM!

PR #372, opened 2023-06-26
Feat(RHIF-199): Add Zero State to Remediations

adonispuente:
Associated Jira ticket: #RHIF-199

Adds zero state when inventory is empty.
# Checklist:

### DIFF
--- a/src/routes/Home.js
+++ b/src/routes/Home.js
@@ -29,6 +29,7 @@ import {
 } from '../hooks/table';
 import ConfirmationDialog from '../components/ConfirmationDialog';
 import keyBy from 'lodash/keyBy';
+import { useChrome } from '@redhat-cloud-services/frontend-components/useChrome';
 
 function verifyDownload(selectedIds, data) {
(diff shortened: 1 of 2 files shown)


src/Routes.js
```diff
@@ -25,25 +30,49 @@ export const routes = {
...
+    try {
+      axios
+        .get(`${INVENTORY_TOTAL_FETCH_URL}?page=1&per_page=1`)
+        .then(({ data }) => {
+          setHasSystems(data.total > 0);
+        });
+    } catch (e) {
+      console.log(e);
+    }
+  }, [hasSystems]);
```
bastilian, line 45: Could the `AppZeroState` component itself not make this request itself?

That way if we’d ever change how we determine when to show the zero state, it would just be in one place to change for all apps.

adonispuente: originally the plan was to make the change in the chrome user obj, but this makes sense too. I'll look into this implementation

PR #378, opened 2023-08-02
Update remediation button RBAC tooltip

leSamo:
https://issues.redhat.com/browse/RHINENG-750

### DIFF
--- a/src/routes/RemediationDetails.js
+++ b/src/routes/RemediationDetails.js
@@ -76,7 +76,7 @@ const RemediationDetails = ({
 
   const context = useContext(PermissionContext);
 
-  const { isFedramp, isBeta } = useChrome();
+  const { isFedramp, isBeta, isOrgAdmin } = useChrome();
   const handleUpsellToggle = () => {
     setUpsellBannerVisible(false);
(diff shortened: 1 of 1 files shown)


bastilian (approved):
LGTM! Thank you @leSamo!

src/routes/RemediationDetails.js
```diff
@@ -94,7 +94,17 @@ const RemediationDetails = ({
 
   const getDisabledStateText = () => {
     if (!context.permissions.execute) {
-      return 'You do not have the required execute permissions to perform this action.';
+      if (isOrgAdmin()) {
```
LiorKGOW, line 97: You could also do something like this:
`Executing the playbook requires having the remediations:remediation:execute permission which is included in the Remediations administrator role. ${isOrgAdmin() ? 'Manage your roles in User access.' : 'Contact your Organization Administrator for access.'}`
Which would remove the string duplication & the if / else statement
What do you think ? @leSamo

LiorKGOW: Sorry for the late response 😅

PR #392, opened 2023-12-12
fix(wepback): update webpack configs to contain deployment tag

adonispuente:
Adds the deployment tag/check to load proper assets in prod and stage envs.


bastilian (approved):
LGTM! Thank you @adonispuente!

PR #394, opened 2024-01-12
fix(RHINENG-3750): Communicate in UI when remediations is disabled

adonispuente:
# Description

Associated Jira ticket: # ([issue](https://issues.redhat.com/browse/RHINENG-3750))

This adds a new endpoint and checks the connection status of the remediation. if the connection status is anything but 'connected', it should be disabled

# How to test the PR

- Open the network tab : Bring down the PR and select a remediation
- in the network tab, filter the results with connection_status
- Open the data and look through the data object for a connection_status key/value pair. If the value is connected (and other conditionals are met) the execute button should be enabled,
- Select multiple remediations, if any of them have a status other than 'connected' and the button is available, then this PR needs more work.
# Before the change
- You could remediate when the connection_status was not 'connected'

# After the change
- You can only execute with this new conditional

### DIFF
--- a/src/Utilities/useConnectionStatus.js
+++ b/src/Utilities/useConnectionStatus.js
@@ -1,0 +1,45 @@
+import { useAxiosWithPlatformInterceptors } from '@redhat-cloud-services/frontend-components-utilities/interceptors';
+const { API_BASE } = require('../config');
+import { useState, useEffect } from 'react';
+
+export const useConnectionStatus = (
+  remediation,
+  context,
+  executable,
+  isFedramp
+) => {
+  const axios = useAxiosWithPlatformInterceptors();
+  const [isDisabled, setIsDisabled] = useState(true);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const connection_status =
+          remediation &&
+          (await axios.get(
+            `${API_BASE}/remediations/${remediation.id}/connection_status`
+          ));
+        let connected =
+          connection_status.data.length > 0
+            ? connection_status.data[0].connection_status !== 'connected'
+            : false;
+        if (!connected) {
+          setIsDisabled(true);
+        } else {
+          setIsDisabled(
+            !context.permissions.execute ||
+              !executable ||
+              isFedramp ||
+              connected
+          );
+        }
+      } catch (error) {
+        console.error(error);
+      }
+    };
+
+    fetchData();
+  }, [remediation, context, executable, isFedramp]);
+
+  return isDisabled;
+};
--- a/src/Utilities/useConnectionStatus.js
+++ b/src/Utilities/useConnectionStatus.js
@@ -1,12 +1,14 @@
 import { useAxiosWithPlatformInterceptors } from '@redhat-cloud-services/frontend-components-utilities/interceptors';
 const { API_BASE } = require('../config');
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 
 export const useConnectionStatus = (remediation) => {
   const axios = useAxiosWithPlatformInterceptors();
   const [isConnected, setisConnected] = useState();
+  const mounted = useRef(false);
 
   useEffect(() => {
+    mounted.current = true;
     const fetchData = async () => {
       try {
         const connection_status =
@@ -14,15 +16,19 @@ export const useConnectionStatus = (remediation) => {
           (await axios.get(
             `${API_BASE}/remediations/${remediation.id}/connection_status`
           ));
-        setisConnected(
-          connection_status.data?.[0].connection_status === 'connected'
-        );
+        mounted.current &&
+          setisConnected(
+            connection_status.data?.[0].connection_status === 'connected'
+          );
       } catch (error) {
         console.error(error);
       }
     };
 
     fetchData();
+    return () => {
+      mounted.current = false;
+    };
   }, [remediation]);
 
   return isConnected;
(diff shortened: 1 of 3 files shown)


src/routes/RemediationDetails.js
```diff
@@ -19,6 +25,7 @@ import ActivityTabUpsell from '../components/EmptyStates/ActivityTabUpsell';
 import DeniedState from '../components/DeniedState';
 import SkeletonTable from '../skeletons/SkeletonTable';
 import '../components/Status.scss';
+import instance from '@redhat-cloud-services/frontend-components-utilities/interceptors';
```
bastilian, line 28: Please use the axios instance via the `useAxiosWithInterceptors` hook.

src/routes/RemediationDetails.js
```diff
@@ -170,6 +179,27 @@ const RemediationDetails = ({
 
   const { status, remediation } = selectedRemediation;
 
+  const fetchDisabledValue = useCallback(async () => {
+    return Promise.resolve(
```
bastilian, line 183: Is the `Promise.resolve` really needed here? `axios.get` should already be an async function and return a promise.

src/routes/RemediationDetails.js
```diff
@@ -170,6 +179,27 @@ const RemediationDetails = ({
...
+      )
+    ).then(({ data }) => {
+      setIsDisabled(
+        !context.permissions.execute ||
+          !executable ||
+          isFedramp ||
+          data[0].connection_status !== 'connected'
+      );
+    });
+  });
```
bastilian, line 195: This should have `context`, `executable` and `isFedramp` as dependencies. 

It might be good to extract the fetching into a hook. something like `useConnectionStatus` that returns only the result of `data[0].connection_status !== 'connected'`

src/routes/RemediationDetails.js
```diff
@@ -170,6 +179,27 @@ const RemediationDetails = ({
...
+
+  useEffect(() => {
+    if (remediation) {
+      fetchDisabledValue();
+    }
+  }, [fetchDisabledValue]);
```
bastilian, line 201: Here `remediation` should be a dependency as well.

src/routes/RemediationDetails.js
bastilian: Extracting it into a hook will also make it easier to add a test for. (😉)

src/routes/RemediationDetails.js
bastilian: Reason for this being that the hook will use the chrome instance coming from the `useChrome` hook. 

This `instance` here will try to access chrome stuff via the `window.insights` object, which we should not do.

src/routes/RemediationDetails.js
adonispuente: the promise doesnt get fulfilled, so when trying to access the value, its empty. It runs too quickly

bastilian: "It runs too quickly".... why does it run too quickly?

src/routes/RemediationDetails.js
adonispuente: Originally I did this, but I had some issues where the other variables would affect things in a weird way, and even if I returned true (for disabling it), it wouldnt update the button. The new changes update everything to be a hook, but it returns a isDisabled value and we pass in those other variables as params

bastilian:
> Attention: 25 lines in your changes are missing coverage. Please review.

😉

src/routes/RemediationDetails.js
bastilian: The hook looks good to me, but if you want we could take a look together to make it purely concerned about the connection status fetching.

src/Utilities/useConnectionStatus.js
```diff
@@ -0,0 +1,29 @@
+import { useAxiosWithPlatformInterceptors } from '@redhat-cloud-services/frontend-components-utilities/interceptors';
+const { API_BASE } = require('../config');
+import { useState, useEffect } from 'react';
+
+export const useConnectionStatus = (remediation) => {
+  const axios = useAxiosWithPlatformInterceptors();
+  const [isConnected, setisConnected] = useState();
+
+  useEffect(() => {
```
bastilian, line 9: To be safe and not cause any unexpected errors or behaviours it might be good if we add a similar guard as here[0], preventing the state update when it's unmounted.

The gist of it is to have a Ref (`mounted` in the example) with a boolean that is set to false by default, set to true in the beginning of the `useEffect` and set to false in the cleanup function. 

[0] https://github.com/RedHatInsights/patchman-ui/pull/1153/files

src/Utilities/useConnectionStatus.js
```diff
@@ -22,7 +24,10 @@ export const useConnectionStatus = (remediation) => {
       }
     };
 
-    fetchData();
+    mounted.current && fetchData();
```
bastilian, line 27: It's rather the `setisConnected` on line 19 that would need to be prevented in case the component is unmounted while the fetch is happening.

adonispuente: ahh I see. Changes are up

bastilian (approved):
Nice work! Thank you @adonispuente!

src/routes/RemediationDetails.test.js
```diff
@@ -0,0 +1,56 @@
...
+    }));
+    const { result, waitForNextUpdate } = renderHook(() =>
+      useConnectionStatus(remediation)
+    );
+
+    await waitForNextUpdate();
+    expect(result.current).toBe(false);
+  });
+
+  //   test('returns false if connection is false', async () => {
```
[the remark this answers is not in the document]
adonispuente: woops, thank you!

PR #405, opened 2024-02-26
Set up fec config

mkholjuraev:
# Description

Associated Jira ticket: # [(issue)](https://issues.redhat.com/browse/RHINENG-7975)

This introduces the new fec.config file together with the ts-loader.

# How to test the PR
1. Try running the app in local env
2. Make sure all config changes are valid

# Before the change

# After the change

# Dependent work link

# Checklist:


package.json
leSamo, line 1: I believe we should update the build to `fec build` as well

README.md
```diff
@@ -8,62 +8,21 @@ This is the frontend application for [Insights Remediations](https://github.com/
...
-2. add `env` attribute with required value to the `webpackProxy` object.
-
-For example, for prod-beta add following:
-```js
-{
-  env: 'prod-beta'
-}
-```
- and run `npm run start:proxy:beta` (the `:beta` suffix configures webpack to serve assets at /preview location.)
+2. Open browser at [https://stage.foo.redhat.com:1337/insights/remediations]( https://stage.foo.redhat.com:1337/insights/remediations)
```
leSamo, line 19: Or possibly at URL with `/preview`

package.json
mkholjuraev: Good catch! will push a change

README.md
mkholjuraev: Makes sense. As the `fec dev` generates a dynamic URL, I have made the wording more abstract, but useful

bastilian:
@mkholjuraev I think the `entry.js`, `entry-dev.js` as well as `bootstrap.js` and `bootstrap-dev.js` could also be removed in the process of upgrading. These have become obsolete and only the files referenced as the `./RootApp` are required for chrome to properly load the applications.

package.json
```diff
@@ -107,31 +108,30 @@
...
     "lint:sass": "stylelint 'src/**/*.scss' --config .stylelintrc.json",
     "server:ctr": "node src/server/generateServerKey.js",
-    "start": "NODE_ENV=development webpack serve --config config/dev.webpack.config.js",
-    "start:beta": "NODE_ENV=development BETA=true webpack serve --config config/dev.webpack.config.js",
-    "start:proxy": "PROXY=true npm run start",
-    "start:proxy:beta": "BETA=true npm run start:proxy",
-    "travis:build": "NODE_ENV=production webpack --config config/test.webpack.config.js",
-    "travis:verify": "npm-run-all build:prod lint test:ct",
+    "start": "PROXY=true HOT=true fec dev",
+    "start:linking": "PROXY=false port=8003 HOT=true fec dev",
```
bastilian, line 127: Other apps use just `npm run start` to only start the web pack devserver without proxy functionality and `npm run start:proxy` to start it with proxy functionality. We should keep it that way.

bastilian: We could make hot reloading default to true in the config and remove it from here.
If someone would want to disable hot reloading they would need to edit the commands or config, because running `HOT=false npm run start` would have no effect, because the `HOT` in the run script will take precedence.

mkholjuraev: Yeah, you are right, especially regarding the consistency with other apps and guidelines. I thought it would be better to make the command shorter. We do not always run the dev server without a proxy. The only case when we use it is when we link the app together and I created a separate readily-available command for it.

mkholjuraev: Will push these changes.

bastilian: Using the `npm run start` to only start the web pack dev server is to be consistent with other/general React/Webpack projects, which makes it easier for others to approach our projects, when they are not familiar with them yet.

mkholjuraev: @bastilian done.

mkholjuraev:
@bastilian Can you please have a look whenever you have a chance?

fec.config.js
```diff
@@ -0,0 +1,40 @@
+const { resolve } = require('path');
+
+module.exports = {
+  appUrl: '/insights/remediations',
+  debug: true,
+  useProxy: process.env.PROXY === 'true',
+  proxyVerbose: true,
+  interceptChromeConfig: false,
+  plugins: [],
+  hotReload: true,
```
bastilian, line 10: What I meant was more like below, like 
```suggestion
  ...(process.env.HOT ? { hotReload: process.env.HOT === 'true' } : { hotReload: true }),
```

bastilian (approved):
LGTM! Thank you @mkholjuraev!

PR #410, opened 2024-03-24
fix(RHINENG-4946): RemediationDetailsTable - Bulk Selector is not updated when an Action is deleted

LiorKGOW:
# Description

This PR addresses issue [#RHINENG-4946](https://issues.redhat.com/browse/RHINENG-4946)

With this PR's changes, we should be able to see that when deleting an Action, the Bulk Selector's number of selected items is being updated

# How to Reproduce the Bug? 
1. Navigate to a remediation that has several Actions
2. Select one or more Actions
3. Click the 'Remove Action' button at the top of the table
4. The Bulk Selector still counts it as selected, although it is no longer present

### DIFF
--- a/src/components/RemediationDetailsTable.js
+++ b/src/components/RemediationDetailsTable.js
@@ -273,7 +273,7 @@ function RemediationDetailsTable(props) {
                   dismissable: true,
                   autoDismiss: true,
                 });
-                selector.reset;
+                selector.reset();
               }}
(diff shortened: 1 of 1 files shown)


AsToNlele (approved):
Thanks for the fix, LGTM :smile:

LiorKGOW:
Thank you for the review @AsToNlele ! 
I have rebased with latest master and repushed 👍🏼

PR #412, opened 2024-04-10
fix(RHINENG-4945): Fix bulk select on archive

johnsonm325:
# Description
Associated Jira ticket: # RHINENG-4953

When a selected playbook is archived, and archived playbooks are not visible, that selection remains in the bulk select.

# How to test the PR
- Navigate to Remediations
- Ensure only non-archived playbooks are being shown via toolbar action kebab
- Select checkbox of a playbook
- Archive playbook from card action kebab

# Before the change
The playbook remains selected (see Bulk Selector).

# After the change
With this PR, when only non-archived playbooks are being shown, and we archive a selected playbook, that selection should not remain.

On the other hand, if we are viewing all playbooks (archived and non-archived) and we archive a selected playbook, the selection should remain.

# Checklist:

### DIFF
--- a/config/setupTests.js
+++ b/config/setupTests.js
@@ -4,7 +4,11 @@ import { enableFetchMocks } from 'jest-fetch-mock';
 enableFetchMocks();
 
 jest.mock('@redhat-cloud-services/frontend-components/useChrome', () => ({
-  useChrome: () => null,
+  useChrome: () => ({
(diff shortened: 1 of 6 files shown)


AsToNlele (approved):
Works exactly as described, LGTM! :smile: 
Thanks for the great description and instructions how to test

PR #415, opened 2024-04-17
feat(RHINENG-9195): Block bootc systems

dkuc:
In this example, localhost is a bootc system.  You can see there is an error and the next button is greyed out. Advisor systems are an exception to this rule. They can remediate.
image

### DIFF
--- a/src/modules/RemediationsModal/steps/reviewSystems.js
+++ b/src/modules/RemediationsModal/steps/reviewSystems.js
@@ -1,5 +1,5 @@
 /* eslint-disable camelcase */
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import propTypes from 'prop-types';
 import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
 import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
Looks reasonable, bit dirty, but it's fine at this point IMO :)

src/modules/RemediationsModal/steps/reviewSystems.js
```diff
@@ -18,18 +18,35 @@ const ReviewSystems = ({ issues, systems, allSystems, registry, ...props }) => {
...
+    const hasBootc = selected.some((s) => bootcIds.includes(s));
+    const isAdvisor = issues?.some((i) => i?.id?.startsWith('advisor'));
+
+    if (hasBootc && !isAdvisor) {
+      setBootcError(
+        'Image mode systems cannot be added to a remediation playbook.'
+      );
+      return input.onChange({});
+    }
+    setBootcError('');
```
bastilian, line 49: This doesn't seem right. Won't this make the `bootcError` always defined and evaluate to truthy below in the render?

dkuc:
> Should we ping team interact for review, or is there still some work left here @dkuc ?

Yes.

src/modules/RemediationsModal/steps/reviewSystems.js
dkuc: Yea it's an empty string which evaluates to false.

mkholjuraev (approved):
LGTM! Works as described.

PR #416, opened 2024-04-23
fix(RHINENG-7065): Refactor popover to include enabled execute button + new api

adonispuente:
# Description

Associated Jira ticket: # ([issue](https://issues.redhat.com/browse/RHINENG-7065))

The popover in remediations details should be enable regardless if the button is disabled or not.
'Connected systems' should only come from the connection_status endpoint where the object actually says 'connected'. if there are not connected systems, the execute button should be disabled. The total amount of systems should also come from this endpoint.

### DIFF
--- a/src/Utilities/useConnectionStatus.js
+++ b/src/Utilities/useConnectionStatus.js
@@ -4,12 +4,13 @@ import { useState, useEffect, useRef } from 'react';
 
 export const useConnectionStatus = (remediation) => {
   const axios = useAxiosWithPlatformInterceptors();
-  const [isConnected, setisConnected] = useState();
+  const [connectedSystems, setConnectedSystems] = useState(0);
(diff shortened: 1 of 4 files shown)


AsToNlele (commented):
Works as described, however shouldn't the `totalSystems` be based on the `system_count` sum and not the `total` in the `meta`?

image

adonispuente:
@AsToNlele I think what youre looking at is the actions tab. Thats saying the CVE affects 2 systems, but not that there are 2 systems associated with that remediation. 
That total number im grabbing uses the meta.total number from the connection_status endpoint, and that shows the total number of systems connected to that remediation.

adonispuente:
changes are up

gkarat (approved):
LGTM! Thanks @adonispuente. I can see the total number of systems is now correct, and the popover is always present

PR #417, opened 2024-04-26
feat(RHINENG:8656): Playbook rename modal doesnt allow duplicate names

adonispuente:
# Description

Associated Jira ticket: # [(issue)](https://issues.redhat.com/browse/RHINENG-8656)

This removes the ability to name a remediation against one that already exist. No duplicates being possible moving forward. (Seperate PR for creation of remediation will be made)

# How to test the PR

grab any remediation name, copy it, and then go to the details of any other remediation.
Then try to rename the remediation with the copied string. The button should be disabled.

I added a timeout that is reset when the user types, so the logic shouldnt perform until the user is done typing (unless they type reaaaaallllly slowly).
While the logic is happening, a spinner takes the buttons place to indicate to the user that something is happening in the background as opposed to it just being disabled.
# Before the change
You could duplicate a remediation with any already-existing name

# After the change
you cannot name a remediation the same as another.

# Dependent work link

# Checklist:

### DIFF
--- a/src/Utilities/useRemediationsList.js
+++ b/src/Utilities/useRemediationsList.js
@@ -12,7 +12,7 @@ export const useRemediationsList = (remediation) => {
     const fetchData = async () => {
       try {
         const nameList = await axios.get(
-          `${API_BASE}/remediations/?fields[data]=${remediation.name}`
+          `${API_BASE}/remediations/?fields[data]=name`
         );
         mounted.current && setRemediationsList(nameList.data);
       } catch (error) {
--- a/src/Utilities/useRemediationsList.js
+++ b/src/Utilities/useRemediationsList.js
@@ -1,0 +1,30 @@
+import { useAxiosWithPlatformInterceptors } from '@redhat-cloud-services/frontend-components-utilities/interceptors';
+const { API_BASE } = require('../config');
+import { useState, useEffect, useRef } from 'react';
+
+export const useRemediationsList = (remediation) => {
+  const axios = useAxiosWithPlatformInterceptors();
+  const [remediationsList, setRemediationsList] = useState();
+  const mounted = useRef(false);
+
+  useEffect(() => {
+    mounted.current = true;
+    const fetchData = async () => {
+      try {
+        const nameList = await axios.get(
+          `${API_BASE}/remediations/?fields[data]=${remediation.name}`
+        );
+        mounted.current && setRemediationsList(nameList.data);
+      } catch (error) {
+        console.error(error);
+      }
+    };
+
+    remediation && fetchData();
+    return () => {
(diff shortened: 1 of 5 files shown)


src/Utilities/useRemediationsList.js
```diff
@@ -0,0 +1,30 @@
...
+  const axios = useAxiosWithPlatformInterceptors();
+  const [remediationsList, setRemediationsList] = useState();
+  const mounted = useRef(false);
+
+  useEffect(() => {
+    mounted.current = true;
+    const fetchData = async () => {
+      try {
+        const nameList = await axios.get(
+          `${API_BASE}/remediations/?fields[data]=${remediation.name}`
```
rexwhite, line 15: This should just be `${API_BASE}/remediations/?fields[data]=${remediation.name}`.  `fields[data]=name` means just return the name field (and nothing else).

src/Utilities/useVerifyName.js
```diff
@@ -0,0 +1,48 @@
...
+    setIsVerifyingName(true);
+
+    //Run a timer 1 second after an input, if the user inputs again within
+    //that timer, clear and reset timer
+    if (timerRef.current) {
+      clearTimeout(timerRef.current);
+    }
+
+    const compareData = async () => {
+      const dataHashmap = {};
```
rexwhite, line 20: A `Set` might be slightly more efficient for this (just a thought, not something that you need to change).

rexwhite (changes requested):

src/Utilities/useRemediationsList.js
adonispuente: ahh i see the difference now, ignore

rexwhite: Yeah, I hit 'save' prematurely 😬.  I fixed my comment...

src/Utilities/useVerifyName.js
adonispuente: This I'd like to keep the same, more so to be consistent with code in other applications.

src/Utilities/useRemediationsList.js
adonispuente: Change is up!

marleystipich2:
Screenshot 2024-04-29 at 10 17 13 AM

I noticed that when you try to update the remediation name to a name that's already being used, the "Save" button is greyed out but there isn't a message explaining why. Was this intentional? I wonder if this would confuse customers?

adonispuente:
@marleystipich2 good catch, i hadnt included it. changes are up

rexwhite (approved):
👍🏻

marleystipich2 (approved):

PR #419, opened 2024-04-30
feat(RHINENG-8657): Creation modal cannot duplicate names + refinement

adonispuente:
# Description

Associated Jira ticket: # ([issue](https://issues.redhat.com/browse/RHINENG-8657))
This PR eliminates the ability to have duplicate names in the creation modal. 
To test run the PR and run apps that consume remediationsButton. For this example ill use advisor.

Advisor -> Recommendations ->Rec Details -> system details ->  select a system and remediate it.
When creating the remediation, you should NOT be allowed to move forward, and be greeted with an error message if
- Remediaton name exist
- Remediation name exist and you add trailing spaces 

Theres no error for an empty remediation but the modal already doesnt allow for this 
# How to test the PR

Refinement work
When editing a remediation name, there should also be errors for the same reasons, and it the remediation name is empty. 
# Dependent work link

# Checklist:

### DIFF
--- a/src/Utilities/useVerifyName.js
+++ b/src/Utilities/useVerifyName.js
@@ -5,6 +5,7 @@ export const useVerifyName = (name, remediationsList) => {
   const [isDisabled, setIsDisabled] = useState(false);
   const mounted = useRef(false);
   const timerRef = useRef(null);
+  const playbookNamePattern = /^(?!$).*[\w\d]+.*$/;
 
   useEffect(() => {
     mounted.current = true;
@@ -17,18 +18,19 @@ export const useVerifyName = (name, remediationsList) => {
     }
 
     const compareData = async () => {
+      const trimmedVal = name.trim();
+
       const dataHashmap = {};
       remediationsList &&
         remediationsList.forEach((item) => {
           dataHashmap[item.name] = true;
         });
 
-      const foundExistingRemediation = (name) => {
-        return dataHashmap[name];
-      };
-      foundExistingRemediation(name)
-        ? setIsDisabled(true)
-        : setIsDisabled(false);
+      if (dataHashmap[trimmedVal] && playbookNamePattern.test(trimmedVal)) {
+        setIsDisabled(true);
+      } else {
+        setIsDisabled(false);
+      }
     };
 
     timerRef.current = setTimeout(() => {
--- a/src/modules/RemediationsModal/steps/selectPlaybook.js
+++ b/src/modules/RemediationsModal/steps/selectPlaybook.js
@@ -25,6 +25,7 @@ import {
   Popover,
   Button,
   Alert,
+  ValidatedOptions,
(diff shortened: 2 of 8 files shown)


src/Utilities/useVerifyName.js
```diff
@@ -5,6 +5,7 @@ export const useVerifyName = (name, remediationsList) => {
   const [isDisabled, setIsDisabled] = useState(false);
   const mounted = useRef(false);
   const timerRef = useRef(null);
+  const playbookNamePattern = /^(?!$).*[\w\d]+.*$/;
```
rexwhite, line 8: The regex used on the api end is `^[^\s].*[^\s]$` - so basically anything as long as it doesn't start or end with whitespace.  Also: \w includes [0-9] so the \d isn't doing anything here.

It's ok if this is more restrictive than the api, though.

src/modules/RemediationsModal/steps/selectPlaybook.js
```diff
@@ -91,6 +93,27 @@ const SelectPlaybook = (props) => {
     }
   });
 
+  //cannot use hook here, as it needs to be called in the OnChange function
+  const verifyName = (val) => {
+    setNewPlaybookName(val);
```
rexwhite, line 98: I'm not following this 100% but it kinda looks like you're setting the new playbook name to the untrimmed version, but comparing the trimmed version to the existing plan names.  If you POST something to the api with leading or trailing spaces in the name you'll get a 400 back though, regardless of whether it's unique or not.

rexwhite (changes requested):
If I'm wrong about my comments let me know and I'll approve this.

src/modules/RemediationsModal/steps/selectPlaybook.js
adonispuente: added a trim to the api request

src/Utilities/useVerifyName.js
adonispuente: this is the pattern baked in from a while back, ill change it to match the api though, i think that makes more sense

src/Utilities/useConnectionStatus.js
```diff
@@ -28,7 +28,9 @@ export const useConnectionStatus = (remediation) => {
           setTotalSystems(totalSystemsCount);
       } catch (error) {
         console.error(error);
-        setDetailsError(error.errors[0].status);
+        setDetailsError(error?.errors[0].status || '');
```
adonispuente, line 31: when the backend returns a 502 error, it doesnt actually send a body in the response. This is just so that if that error is hit, the page doesnt explode trying to locate it. For other errors, a body is returned

rexwhite: We don't need no stinking' body!....   (also: gross, but this will work for now.  I'll add this to the list of things to fix in the api...)

PR #422, opened 2024-05-07
fix(RHINENG-8088): refactor select playbook typeahead

johnsonm325:
# Description

Associated Jira ticket: # 8088

When you are creating playbook and choose from Existing playbooks (Add to existing playbook) select appears without search option. Seems it happened after Patternfly 5 migration. 

# How to test the PR

To repro:
- Choose and click a system in inventory
- From the system details page, click the Vulnerability tab Pick CVE(s) and click on Remediate button
- When modal opens - click in the "Select playbook" box

# Before the change
image

This box should allow typeahead, which filters the existing playbooks as you type. But this feature seems to have been removed from the PF component we were using and have been moved to the Select component.

# After the change
image

This PR refactors this box to allow typeahead.

# Checklist:

- [x] Screenshots before and after the change are added
- [x] Tests for the changes have been added
- [x] README.md is updated if necessary

### DIFF
--- a/src/modules/RemediationsModal/common/ExistingPlaybookTypeahead.js
+++ b/src/modules/RemediationsModal/common/ExistingPlaybookTypeahead.js
@@ -1,0 +1,204 @@
+import React, { useEffect, useRef, useState } from 'react';
+import {
+  Select,
+  SelectOption,
+  SelectList,
+  MenuToggle,
+  TextInputGroup,
+  TextInputGroupMain,
+} from '@patternfly/react-core';
+import propTypes from 'prop-types';
+import { EXISTING_PLAYBOOK } from '../../../Utilities/utils';
+import * as api from '../../../api';
+
+const ExistingPlaybookTypeahead = ({
+  selectedPlaybook,
+  setIsLoadingRemediation,
+  setSelectedPlaybook,
+  existingPlaybookSelected,
+  existingRemediations,
+  input,
+  formOptions,
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const [filterValue, setFilterValue] = useState('');
+  const [selectOptions, setSelectOptions] = useState(existingRemediations);
+  const [focusedItemIndex, setFocusedItemIndex] = useState(null);
+  const [activeItem, setActiveItem] = useState(null);
+  const textInputRef = useRef();
+
+  useEffect(() => {
+    let newSelectOptions = existingRemediations;
+    if (filterValue) {
+      newSelectOptions = existingRemediations.filter((menuItem) =>
+        String(menuItem.name).toLowerCase().includes(filterValue.toLowerCase())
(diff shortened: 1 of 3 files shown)


src/modules/RemediationsModal/common/ExistingPlaybookTypeahead.js
```diff
@@ -0,0 +1,204 @@
...
+    switch (event.key) {
+      case 'Enter':
+        if (isOpen && focusedItem.name !== 'no results') {
+          onSelect(null, focusedItem.id);
+        }
+        setIsOpen((prevIsOpen) => !prevIsOpen);
+        setFocusedItemIndex(null);
+        setActiveItem(null);
+        break;
+      case 'Escape':
```
mkholjuraev, line 123: This is causing the wizard to close instead of the select dropdown. I believe this is unexpected, or?

src/modules/RemediationsModal/common/ExistingPlaybookTypeahead.js
```diff
@@ -0,0 +1,204 @@
...
+  const onInputKeyDown = (event) => {
+    const enabledMenuItems = selectOptions.filter(
+      (option) => !option.isDisabled
+    );
+    const [firstMenuItem] = enabledMenuItems;
+    const focusedItem = focusedItemIndex
+      ? enabledMenuItems[focusedItemIndex]
+      : firstMenuItem;
+    switch (event.key) {
+      case 'Enter':
```
mkholjuraev, line 115: When I search for a non-existing option and press 'Enter', the selection does not end, the loading bar continues spinning, and there is no way to re-select playbooks unless you re-open the wizard

johnsonm325:
@mkholjuraev Take another look. You may want to take a look at a few more test scenarios.

Try typing in the text and pressing enter before you've used the arrow keys to select a playbook.

Also, before this new commit, after you've selected a playbook, it would load the playbook into the select box, but there was no way to type into the text box after the fact.

This commit updates this so after a playbook is selected, you can still edit the text in the box.

mkholjuraev (commented):
I had another look at this and also those new cases. All seem to be fine. The only issue left is ESC key handling. Still, the whole wizard gets closed instead of the select dropdown. Otherwise, looks great!

johnsonm325:
@mkholjuraev I [removed the escape key handling](https://github.com/RedHatInsights/insights-remediations-frontend/pull/422/commits/b7430491a5dde3dd6dfe061e4d526252f302be84#diff-4c09416000727398441b1dc0fe3f2f5f9940d3a2cc01734d47da37370bf29b25L123-L126). Seems that no matter whether it's in there or not, it still closes the entire modal. Even if I include `event.preventDefault()`, it still closes the entire modal.

If you can think of a way to make it only handle the dropdown, then I can include it.

PR #428, opened 2024-08-14
fix: Execution modal show Not available if system isn't connected

leSamo:
https://issues.redhat.com/browse/RHINENG-4955

Wrapped the existing code block inside ternary statement which shows "Not available" if `con.connection_status` isn't 'connected'


adonispuente (commented):
The code itself works, but its not super readable.
Its more of a nit pick, but if the if/else statements are easier to understand, could we replace it with something like this? 

Also, could we add test for this modal


`let title;

  if (con.connection_status === 'connected') {
    if (con.executor_name) {
      const truncatedName =
        con.executor_name.length > 25
          ? `${con.executor_name.slice(0, 22)}...`
          : con.executor_name;
      title = (
        <Tooltip content={`${con.executor_name}`}>
          <span>{truncatedName}</span>
        </Tooltip>
      );
    } else {
      title = 'Direct connection';
    }
  } else {
    title = 'Not available';
  }`

src/components/Modals/ExecuteModal.js
```diff
@@ -66,17 +66,22 @@ export const ExecuteModal = ({
...
+              <Tooltip content={`${con.executor_name}`}>
+                <span>
+                  {con.executor_name.length > 25
+                    ? `${con.executor_name.slice(0, 22)}...`
+                    : con.executor_name}
+                </span>
+              </Tooltip>
+            ) : (
+              'Direct connection'
+            )
```
bastilian, line 81: We should maybe at least put this into a discrete component and make the teneray operations going on here a bit more digestible.

src/components/Modals/ExecuteModal.js
```diff
@@ -66,17 +66,22 @@ export const ExecuteModal = ({
   const rows = [...connected, ...disconnected].map((con) => ({
```
bastilian, line 66: If we would extract the `rows` building into a function we could add a test for this change, without having to go through setting up tests for the whole `ExecuteModal` component.

src/components/Modals/ExecuteModal.js
leSamo: I separated it into a function

src/components/Modals/ExecuteModal.js
leSamo: Now that I've separated it into a function, it's now much more readable, and does not warrant a test IMO

bastilian: We should start to increase the coverage in remediation. Please add a test.

bastilian: If we extract the whole rows building into a function outside the component and add a test for it, one asserting being it generates the right status it is a start to improve the coverage with just a tiny bit more effort.

leSamo: I added a test which is a bit more E2E, is it okay?

bastilian (approved):
LGTM! Thank you @leSamo!

PR #429, opened 2024-08-14
Delete obsolete PlanModal components

leSamo:
These components have not been touched in 3+ years and have no use inside or outside of the Remediation application. I suppose they are a part of feature that was cut.


AsToNlele (approved):
makes sense, LGTM 😄

PR #430, opened 2024-08-15
Refactor DeleteButton to a functional component 

leSamo:


src/skeletons/SkeletonTable.js
```diff
@@ -8,55 +8,47 @@ import { EmptyTable } from '@redhat-cloud-services/frontend-components/EmptyTabl
...
-          { title: <Skeleton size="md" /> },
-          { title: <Skeleton size="xs" /> },
-          { title: <Skeleton size="xs" /> },
-          { title: <Skeleton size="md" /> },
-          { title: <Skeleton size="md" /> },
-        ],
-      ],
-    };
-  }
+const SkeletonTable = ({ ...props }) => {
```
bastilian, line 11: Could we maybe use the `SkeletonTable` component from the component-groups package and just drop this component?

leSamo: You are right, I removed the `SkeletonTable` refactor from this PR and I will create another PR where I delete the `SkeletonTable`

AsToNlele (approved):
Works the same, LGTM

PR #432, opened 2024-08-19
Remove skeleton table

leSamo:
This change migrates from custom SkeletonTable which didn't look that good to a better version from `@patternfly/react-component-groups`, this saves some code and improves the UXD by adding table headers to the SkeletonTable.


bastilian (approved):
Nice! Thank you @leSamo!

PR #433, opened 2024-08-27
Reset  filter after Playbook deletion

leSamo:
https://issues.redhat.com/browse/RHINENG-4954

Before -- after deleting the playbook the filter did not reset, but the cards below were fetched as if the filter wasn't applied
After -- after deleting the playbook the filter is reset and cards are shown correctly


AsToNlele (approved):
Thanks for the fix, LGTM

PR #441, opened 2024-10-03
fix(RHINENG-11503): Playbook log doesn't display properly

johnsonm325:
Playbook log doesn't display properly after inital load.

To repro:
- Automation Toolkit > Remediations > A playbook  > Activity > The data that the playbook ran successfully > Direct connected
- At this point the playbook log will flash on screen and disappear
- Click on the hostname

The issue was that playbookRunSystemDetails is a redux variable that remains after initial load. So when the playbook log is loaded again, the log viewer somehow only captures part of the string or none at all.

The best way to handle this issue is to clear the variable in redux on inital load to ensure there are no incorrect states.

### DIFF
--- a/src/components/ExecutorDetails/ExecutorDetails.js
+++ b/src/components/ExecutorDetails/ExecutorDetails.js
@@ -1,6 +1,5 @@
 import React, { useEffect, useState, useRef, useContext } from 'react';
-import { connect } from 'react-redux';
-import PropTypes from 'prop-types';
+import { useDispatch, useSelector } from 'react-redux';
 import Link from '@redhat-cloud-services/frontend-components/InsightsLink';
 import { useParams } from 'react-router-dom';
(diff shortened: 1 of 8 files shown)


src/components/ExecutorDetails/ExecutorDetails.js
```diff
@@ -46,25 +45,34 @@ import columns from './Columns';
 import { useGetEntities } from './helpers';
 
 let refreshInterval;
+const _ = require('lodash');
```
bastilian, line 48: Since we only need `isEmpty` we could use `import isEmpty from 'lodash/isEmpty'`.

AsToNlele:
@johnsonm325 do you have an account I could test this on?

johnsonm325:
@AsToNlele It does not have to be a playbook that has a successful run. It just needs to have a playbook log.

AsToNlele (approved):
lgtm, thanks for writing a test! 😄

PR #455, opened 2024-12-05
fix(RHINENG-9779): Change Details Page Header layout

adonispuente:
https://issues.redhat.com/browse/RHINENG-9779

Alters layout of remediation details page header so that if theres a really long name it doesnt move the buttons, but instead just wraps under itself:

Before PR: 
Screenshot 2024-12-05 at 1 32 17 PM

With PR 
Screenshot 2024-12-05 at 1 31 23 PM


leSamo (approved):
LGTM!

PR #456, opened 2025-01-02
fix(Remediation wizard): Fix incorrect system count in the pagination

leSamo:
Fixes https://issues.redhat.com/browse/RHINENG-10618

- Remove unnecessary object construction with instant deconstruction
- Do not overwrite `total` property of the spread `data` object -- it contains the correct system count since `total` property includes row count which is double due to including expand cells as rows


adonispuente (approved):
works as intended, LGTM

PR #480, opened 2025-03-14
Implement selection, page select, and sorting on overviewPage

adonispuente:
# Description

Associated Jira ticket: # (issue)
[RHINENG-16094](https://issues.redhat.com/browse/RHINENG-16094) 
and partially [RHINENG-16095](https://issues.redhat.com/browse/RHINENG-16094)

In the overview page you should be able to sort on each column -> some options not supported by the BE but the api request should look valid

You can also select items on the list

# How to test the PR

Run the PR on stage env, FF is already on, its on the home page so as so as you navigate to the app youll see the table. Just start clicking the columns and verify the table and request indicate sorted data.

### DIFF
--- a/src/Frameworks/AsyncTableTools/AsyncTableTools/hooks/useTableSort/useTableSort.js
+++ b/src/Frameworks/AsyncTableTools/AsyncTableTools/hooks/useTableSort/useTableSort.js
@@ -56,7 +56,7 @@ const useTableSort = (columns, options = {}) => {
   const [sortBy, setSortBy] = useTableState(
     TABLE_STATE_NAMESPACE,
     initialSortBy || {
-      index: 1,
+      index: 0,
       direction: 'asc',
     },
     stateOptions
--- a/src/Frameworks/AsyncTableTools/AsyncTableTools/hooks/useTableSort/useTableSort.js
+++ b/src/Frameworks/AsyncTableTools/AsyncTableTools/hooks/useTableSort/useTableSort.js
@@ -56,7 +56,7 @@ const useTableSort = (columns, options = {}) => {
   const [sortBy, setSortBy] = useTableState(
     TABLE_STATE_NAMESPACE,
     initialSortBy || {
-      index: 0,
+      index: 1,
       direction: 'asc',
     },
     stateOptions
@@ -77,7 +77,6 @@ const useTableSort = (columns, options = {}) => {
     ...sortBy,
     index: sortBy?.index + offset,
   };
-
   return {
     tableProps: {
       onSort,
--- a/src/Utilities/Hooks/useQuery.js
+++ b/src/Utilities/Hooks/useQuery.js
@@ -1,5 +1,6 @@
 import { useRef, useState } from 'react';
 import { useDeepCompareEffect, useDeepCompareCallback } from 'use-deep-compare';
+import { fetchResult } from './api/helpers';
 
 /**
  *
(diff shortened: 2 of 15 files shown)


src/Frameworks/AsyncTableTools/AsyncTableTools/hooks/useTableSort/useTableSort.js
```diff
@@ -34,7 +34,7 @@ import { TABLE_STATE_NAMESPACE } from './constants';
  *
  */
 const useTableSort = (columns, options = {}) => {
-  const { sortBy: initialSortBy, serialisers, onSort: onSortOption } = options;
+  const { filter: initialSortBy, serialisers, onSort: onSortOption } = options;
```
bastilian, line 37: let's try and not do that. :D

src/api/useRemediationsQuery.js
```diff
@@ -97,6 +98,7 @@ const useRemediationsQuery = (
           error: queryError,
           loading: queryLoading,
           meta: metaData,
+          availableIDs: availableIDs,
```
bastilian, line 101: Line 85 is what you are looking for. ;) The `fetchAllIds` function can be passed directly to the Table options as `itemIdsInTable`.

src/routes/OverViewPage.js
```diff
@@ -51,12 +51,15 @@ export const OverViewPage = () => {
...
+          index: 6,
+          direction: 'desc',
+        },
         itemIdsInTable: fetchAllIds,
         manageColumns: true,
         onSelect: handleSelectionChange,
-        itemIdsOnPage: data,
-        identifier: 'id',
-        preselected: data && data[0]?.id,
+        itemIdsOnPage: availableIDs,
```
bastilian, line 61: Just do `itemIdsOnPage: data.map(({id}) => id)`. No `availableIDs` thingy needed. :)

src/api/useRemediationsQuery.js
bastilian: oh, now I get what you want.

src/Utilities/Hooks/useQuery.js
```diff
@@ -45,11 +45,9 @@ const useQuery = (fn, options = {}) => {
   const [meta, setMeta] = useState(undefined);
   const [error, setError] = useState(undefined);
   const [loading, setLoading] = useState(false);
-  console.log(params, 'params outside useQuery here');
-
+  const [availableIDs, setAvailableIDs] = useState([]);
```
bastilian, line 48: Remember what I said earlier today. The `useQuery` should only be concerned with calling a function and returning the results either as a state or directly.

It is not meant to alter, separate out or do anything with the result.
It shouldn't know about `meta` either. These are things that if anything the `useRemediationsQuery` should be responsible for.

bastilian:
@adonispuente I'd recommend to grab the `useQuery` & `useFetchTotalBatched` from compliance (once this PR: https://github.com/RedHatInsights/compliance-frontend/pull/2433 is merged, or grab it from there) and let's try to use the `compileResult` feature to get a consistent format or the responses results. You probably can just grab the `compileResults` function from compliance as well and it should theoretically work. 

You can then have your api hook calls be like this

```js 
const { result, loading, error } = useRemediationsQuery(
```

with the `result` being `{ data: [...], meta: {} }`. always.

src/Frameworks/AsyncTableTools/AsyncTableTools/hooks/useTableSort/useTableSort.js
```diff
@@ -56,7 +56,7 @@ const useTableSort = (columns, options = {}) => {
   const [sortBy, setSortBy] = useTableState(
     TABLE_STATE_NAMESPACE,
     initialSortBy || {
-      index: 0,
+      index: 1,
```
bastilian, line 59: I thin this can be discarded. I think I typed `1`, because I forgot for a second indexes start with 0...

src/Utilities/Hooks/useQuery.js
adonispuente: Ahh youre right, I was about to change it to also just bring back data, instead of data.data, and then use meta with data.meta, but im gonna go ahead with the compileResult thing instead

src/Frameworks/AsyncTableTools/AsyncTableTools/hooks/useTableSort/useTableSort.js
adonispuente: I'll slap this up too

adonispuente:
@bastilian Thanks for all the feedback, I just sent up all the improvements!

bastilian (approved):
LGTM! Thank you @adonispuente!

PR #495, opened 2025-04-16
fix(RHINENG:16087): Small flow flixes

adonispuente:
[RHINENG-16087](https://issues.redhat.com/browse/RHINENG-16087)

This fixes a few issues found in QA

-User cannot do any bulk action filters if nothing is selected : 
Screenshot 2025-04-16 at 12 13 14 PM

There was a renaming error where we wouldnt refetch one a name is changed. Now when you rename an item, the remList will be updated, so when you go to rename an additional item, you cannot match the name of a previous one: example flow 

If I do

take a playbook, rename it to "abc" (choose a non-existent name)
rename a different playbook to "abc"
The modal displays no error. Instead we get the error in an alert. The renaming doesn't happen.
If I do
take a playbook named "abc"
rename it to "abcde"
try to rename it back to "abc"

Now, the modal will tell you that you can rename something to a playbook with an existing name. To test just follow that flow

### DIFF
--- a/src/components/ConfirmationDialog.js
+++ b/src/components/ConfirmationDialog.js
@@ -27,7 +27,7 @@ export default function ConfirmationDialog({
           variant="danger"
           ouiaId="confirm"
           onClick={() => onClose(true)}
-          isDisabled={selectedItems && selectedItems?.length === 0}
+          isDisabled={selectedItems?.length === 0}
         >
           {confirmText}
         </Button>,
--- a/src/components/ConfirmationDialog.js
+++ b/src/components/ConfirmationDialog.js
@@ -10,6 +10,7 @@ export default function ConfirmationDialog({
   text = 'This action cannot be undone',
   confirmText = 'Remove system',
   onClose = (f) => f,
+  selectedItems,
 }) {
   return (
     <Modal
@@ -26,6 +27,7 @@ export default function ConfirmationDialog({
           variant="danger"
           ouiaId="confirm"
           onClick={() => onClose(true)}
+          isDisabled={selectedItems && selectedItems?.length === 0}
         >
           {confirmText}
         </Button>,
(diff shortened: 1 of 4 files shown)


src/components/ConfirmationDialog.js
```diff
@@ -26,6 +27,7 @@ export default function ConfirmationDialog({
           variant="danger"
           ouiaId="confirm"
           onClick={() => onClose(true)}
+          isDisabled={selectedItems && selectedItems?.length === 0}
```
leSamo, line 30: This could be shortened to:

`isDisabled={selectedItems?.length === 0}`

Since the optional chaining takes cares of the case when selectedItems is 0.

src/components/NoResultsTable.js
```diff
@@ -26,7 +28,13 @@ export const NoResultsTable = () => {
         infrastructure.
       </EmptyStateBody>
       {/*TODO:  Need to link this somewhere */}
```
leSamo, line 30: I think this TODO has been fulfilled

src/routes/OverViewPage/OverViewPage.js
```diff
@@ -215,20 +217,32 @@ export const OverViewPage = () => {
...
       {
-        label: (
-          <TextContent className="pf-v5-u-danger-color-100">Delete</TextContent>
-        ),
+        label: 'Delete',
+        props: {
+          className:
+            !context.permissions.write || !currentlySelected?.length
+              ? 'var(--pf-v5-global--disabled-color--200)'
+              : 'pf-v5-u-danger-color-100',
```
leSamo, line 242: This is a bit inconsistent -- once using "var", second time not

leSamo (commented):
Thanks for the PR -- functionally good, just some code clarity nitpicks

adonispuente:
@leSamo Thanks for the quick review and catching those, I just sent those changes up now! Theres also some time reserved once all functionality is implemented for refactoring/deleting old code as a lot of it is super old and unneeded

leSamo (approved):
Looks great!

PR #498, opened 2025-04-22
feat(RHINENG-16089): Implement Actions Tab for Remediations Details Remodel

adonispuente:
This ticket implements: https://issues.redhat.com/browse/RHINENG-16089 

Unsure if I can have mocks directly here so just incase: https://www.figma.com/design/FyfDGBiNpa1CgArgJd9bHk/CPUX5904_remediation_redoMVP?node-id=8037-12817&p=f&t=KW0LAgPgLlgn2C02-0 

To test this PR, go to a remediations details actions tab. You should be able to perform:
Pagination
Filter by name
Bulk Select + Bulk remove actions
click the systems column item to open a modal
remove an individual action 

**Sorting** is temporarily disabled until BE makes some updates


src/routes/RemediationDetailsComponents/helpers.js
```diff
@@ -75,3 +75,15 @@ export const getTimeAgo = (date) => {
   const diffYears = Math.floor(diffDays / 365);
   return `${diffYears} year${diffYears === 1 ? '' : 's'} ago`;
 };
+
+// eslint-disable-next-line react/display-name
+export const renderComponent = (Component, props) => (_data, _id, entity) =>
+  <Component {...entity} {...props} />;
+
+export const chunkArray = (array, size) => {
```
leSamo, line 83: I believe using lodash function `chunk` could save some code here: https://www.geeksforgeeks.org/lodash-_-chunk-method/

leSamo (commented):
I see you are using here `||` short circuiting quite a lot, maybe you'd want to rather use [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) `??` to increase code clarity and prevent bug when falsish, yet non-nullish value (like integer `0`) unintentinally trigger short circuiting when not intended.

leSamo (commented):
I also believe adding `description: ''` to `dispatchNotification` is redundant and the line can be removed.

src/routes/RemediationDetailsComponents/ActionsContent.js
```diff
@@ -0,0 +1,209 @@
...
+        aria-label="ActionsTable"
+        ouiaId="ActionsTable"
+        items={pageOfIssues}
+        total={filteredIssues.length}
+        columns={[...columnsWithSystemsButton]}
+        filters={{
+          filterConfig: [...actionNameFilter],
+        }}
+        options={{
+          onSelect: () => '',
```
leSamo, line 161: Why would `onSelect` return a string here?

adonispuente: Ill add a note because without context this is very odd, this is a known issue in AsyncTableTools, and will be fixed once the PR for it is merged and we convert a lot of the code here to be used from the new AsyncTableTools package

src/routes/RemediationDetailsComponents/helpers.js
adonispuente: ahh this is nice, added!

johnsonm325 (approved):

PR #499, opened 2025-04-23
feat(RHINENG-16090): Implement Remodeled Details System Tab

adonispuente:
This PR implements the Systems Tab in the remodeled table. Currently this mostly includes work for when we import a new table using asyncTableToolsHooks. It includes prep work for that as well as actually refactoring/implementing the current Systems Table in the systems tab. Once AsyncTableTools is updated, we will continue the migration to SystemsContent


src/routes/RemediationDetailsComponents/SystemsContent/Columns.js
```diff
@@ -0,0 +1,55 @@
...
+  RebootRequiredCell,
+  OSCell,
+  ConnectionStatusCell,
+} from './Cells';
+
+export default [
+  {
+    title: 'Name',
+    transforms: [wrappable],
+    // sortable: 'action',
```
gkarat, line 16: Are we planning to enable sorting? What are the current constraints?

gkarat (approved):
I couldn't find any issues, the code looks good. Let's finish the migration to SystemContent in a separate PR and then make final test and review on this component.

src/routes/RemediationDetailsComponents/SystemsContent/Columns.js
adonispuente: Currently, in the old version of things the FE does all the sorting, there is work being done on the BE to imlpement sorting properly, once thats added, we should just be able to uncomment these bits in the columns and sorting should just work

PR #500, opened 2025-04-23
fix(emptyPage): Update empty page with padding and making it full size

adonispuente:
This updates emptyState (no remediations) so that it extends the full page.
To test, update the conditional in OverViewPage to say
 `allRemediations?.data.length > 0 ? `instead of `allRemediations?.data.length === 0 ?`

OR log in with an account that has no remediations  
PR:
Screenshot 2025-04-23 at 12 29 07 PM

Mocks:
Screenshot 2025-04-23 at 12 27 51 PM


LightOfHeaven1994 (commented):
@adonispuente there are 2 things:

1) Empty state is not shown if you have 0 active playbooks but you have archived. I checked current implementation on prod and I see API call for empty state has `hide_archived: true`. But empty state is different as it still shows toolbar where you can unarchive playbooks. Screenshot shows old empty state.
IDK how handle it better if you have only archived playbooks. We probably need to double check if with UXD
image


2) I see that we have 2 almost same links and one of them should be removed (it's not existing on mocks):
image

Other than that - PR looks great 😄

PR #525, opened 2025-05-27
Fix(MVP): Fix small bugs and update execution flow 

adonispuente:
There are no Jira tickets associated with the following bugs, these were just found. 

The tooltip over execution button now ONLY appears when the button is disabled
Passed down the refetchRem function so rename on the details card works
Update the flow on the execution button so when a user executes the section reloads (This whole component is getting rewritten in another PR)

## Summary by Sourcery

Fix small bugs in execution button UI and refetch logic, and update the execution flow to use the new executeRemediation API with remediationsV2 feature flag and data refetch on completion.

Bug Fixes:
- Display tooltip over execute button only when it’s disabled
- Pass refetchRemediationPlaybookRuns prop through detail components to enable rename refresh

Enhancements:
- Centralize execution logic in ExecuteModal.handleClick supporting both legacy and v2 flows via remediationsV2 flag
- Introduce executeRemediation API call and integrate useRemediationsQuery for playbook runs
- Refresh remediation playbook runs and close modal on successful execution

### DIFF
--- a/src/components/ExecuteButton.js
+++ b/src/components/ExecuteButton.js
@@ -36,38 +36,42 @@ const ExecuteButton = ({
   }, [remediationStatus]);
 
   const buttonWithTooltip = () => {
-    return (
+    const button = (
+      <Button
+        isAriaDisabled={isDisabled}
+        data-testid="execute-button-enabled"
+        onClick={() => {
+          setOpen(true);
+          getConnectionStatus(remediation.id);
+        }}
+      >
+        {isDisabled && <ExclamationTriangleIcon />} Execute
+      </Button>
+    );
+
+    return isDisabled ? (
       <Tooltip
         minWidth="400px"
         aria-label="details Tooltip"
         content={
-          <>
-            <Flex
-              className="pf-v5-u-ml-md"
-              direction={{ default: 'column' }}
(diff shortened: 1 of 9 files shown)


src/components/ExecuteButton.js
```diff
@@ -90,6 +95,7 @@ const ExecuteButton = ({
           issueCount={issueCount}
           runRemediation={runRemediation}
```
LightOfHeaven1994, line 96: We can remove `runRemediation` from here since ExecuteModal has no such prop anymore

src/components/Modals/ExecuteModal.js
```diff
@@ -308,4 +332,5 @@ ExecuteModal.propTypes = {
   runRemediation: PropTypes.func,
```
LightOfHeaven1994, line 332: We can remove `runRemediation` from here since ExecuteModal has no such prop anymore

src/components/ExecuteButton.js
LightOfHeaven1994: And also from propTypes on line 105

LightOfHeaven1994 (approved):
Generally looks good, but we can do a small cleanup (also remove `runRemediation` from execute modal test)

PR #528, opened 2025-06-03
feat(RHINENG): 18471: Add refresh button to exectHist Tab

adonispuente:
https://issues.redhat.com/browse/RHINENG-18471 

This implements a refresh button on the execution history tab.
There arent any mocks, but the functionality is as follows :

There is a refresh button on the execution history tab (details page) on the right side of the screen.
When clicked, it should reload the tables and disable the refresh button while its loading
It should also be a skeleton table while its loading. 
You should be able to move around to other tabs while its loading

Screenshot 2025-06-04 at 11 23 51 AM
Screenshot 2025-06-04 at 11 24 49 AM

## Summary by Sourcery

Enable manual refreshing of the execution history tab by adding a "Refresh" button that reloads the run systems table with a skeleton loader while preserving tab navigation.

New Features:
- Add an inline Refresh button with a RedoIcon in the Execution History run tab to trigger a manual reload.

Enhancements:
- Combine manual refresh state with existing loading state to display a skeleton table instead of a spinner during reload.
- Prevent the Refresh button from appearing while loading to avoid duplicate requests.
- Update RunSystemsTable to forward the loading prop to the RemediationsTable component.


hugohezel (approved):
**Code**: Looks good to me.
**Functionality**: Refresh button works well and reliably. I really like reasonable saving on resources -  after refreshing runs, opening runs that were opened before the refresh doesn't create new requests for systems.
**Visuals**: One small nit - there is no padding between the refresh button and the right edge, which feels inconsistent with the rest of the layout.

adonispuente:
@hugohezel In regards to visuals, I actually thought the same thing. For the time being though UX wants to button as is.

Thank you for reviewing!

PR #540, opened 2025-06-27
refactor(RHINENG-18944): Migrate to ATT package

adonispuente:
https://issues.redhat.com/browse/RHINENG-18944

All behavior in the app should behave the same. Nothing should be different in behavior from master branch. For testing simply try to use the app as normal and hit any edge case you can think of

### DIFF
--- a/src/routes/OverViewPage/CalendarFilterType.js
+++ b/src/routes/OverViewPage/CalendarFilterType.js
@@ -3,13 +3,10 @@ import CalendarFilter from '../CalendarFilter';
 export const CalendarFilterType = {
   Component: CalendarFilter,
   chips: (value = []) => {
-    console.log(value, 'value here');
     return value.length ? [value[0].slice(0, 10)] : [];
   },
 
   selectValue: (selectedValue) => {
-    console.log(selectedValue, 'selectedValue here');
-
     return [[selectedValue], true];
   },
   deselectValue: () => [undefined, true],
--- a/src/routes/OverViewPage/OverViewPage.js
+++ b/src/routes/OverViewPage/OverViewPage.js
@@ -184,7 +184,7 @@ export const OverViewPage = () => {
               ouiaId="OverViewTable"
               variant="compact"
               loading={loading}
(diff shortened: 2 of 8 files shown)


src/routes/OverViewPage/CalendarFilterType.js
```diff
@@ -0,0 +1,16 @@
+import CalendarFilter from '../CalendarFilter';
+
+export const CalendarFilterType = {
+  Component: CalendarFilter,
+  chips: (value = []) => {
+    console.log(value, 'value here');
```
bastilian, line 6: 👀

src/routes/OverViewPage/OverViewPage.js
```diff
@@ -181,7 +184,7 @@ export const OverViewPage = () => {
               ouiaId="OverViewTable"
               variant="compact"
               loading={loading}
-              items={result?.data}
+              items={result?.data.map((item) => ({ ...item, itemId: item.id }))}
```
bastilian, line 187: With the latest version of the tabletools this should not be needed anymore.

src/routes/RemediationDetailsComponents/ActionsContent/ActionsContent.js
```diff
@@ -1,23 +1,25 @@
-import React, { useCallback, useMemo, useState } from 'react';
-import RemediationsTable from '../../../components/RemediationsTable/RemediationsTable';
+/* eslint-disable react/prop-types */
```
bastilian, line 1: 🤨 (there is no friendly version of this emoji, it should be a friendly eyebrow raise.)

adonispuente: Component: ({ systems, description }) =>

the props on this line scream, and the proper fix is to move the inline component out of things and name it, but that cause another slew of issues that causes everything to get super ugly

adonispuente: ehh i changed it anyways

PR #549, opened 2025-07-11
fix(RHINENG:18973): delete systems flow fix

adonispuente:
This is a bandaid fix for the inventory table to refresh once you delete a system on the details page. 
https://issues.redhat.com/browse/RHINENG-18973

Currently inventory.current.onRefresh data is only working after the first delete, adding a key prop causes react to rerender the entire table, so the behavior isnt desired but it works. Will implement the table migration to the new inventory table once its complete.

This also increases coverage to 43%
## Summary by Sourcery

Implement a temporary fix to force the inventory table to refresh after deleting systems and refactor table logic into reusable hooks

Bug Fixes:
- Force InventoryTable to remount by adding a dynamic key on remediation changes to ensure proper data refresh after system deletion

Enhancements:
- Extract bulk selection logic into a useBulkSelect hook
- Extract delete confirmation and notification logic into a useOnConfirm hook
- Add normalizeConnectedData helper to streamline connection data mapping in fetchInventoryData
- Refactor SystemsTableWrapper to simplify dispatch usage, use custom hooks, and wire up InventoryTable props consistently

Chores:
- Update RemediationDetails to rename and propagate refetchRemediationDetails callback to child components
- Refactor useConnectionStatus hook signature to accept remediationId and external axios instance, and fix effect dependencies

### DIFF
--- a/src/Utilities/useConnectionStatus.js
+++ b/src/Utilities/useConnectionStatus.js
@@ -1,15 +1,13 @@
 import { useState, useEffect, useRef } from 'react';
 import { API_BASE } from '../routes/api';
-import { useAxiosWithPlatformInterceptors } from '@redhat-cloud-services/frontend-components-utilities/interceptors';
 
-export const useConnectionStatus = (remediationId) => {
+export const useConnectionStatus = (remediationId, axios) => {
   const [connectedSystems, setConnectedSystems] = useState(0);
   const [totalSystems, setTotalSystems] = useState(0);
   const [areDetailsLoading, setAreDetailsLoading] = useState(true);
   const [detailsError, setDetailsError] = useState();
   const [connectedData, setConnectedData] = useState([]);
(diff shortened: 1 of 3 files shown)


src/Utilities/useConnectionStatus.js
```diff
@@ -1,23 +1,22 @@
-import { useAxiosWithPlatformInterceptors } from '@redhat-cloud-services/frontend-components-utilities/interceptors';
+/* eslint-disable react-hooks/exhaustive-deps */
```
bastilian, line 1: 🫣

AsToNlele (commented):
`RemediationDetails.js` is missing the `axios` variable.
```
const axios = useAxiosWithPlatformInterceptors();
```

We can prevent this by adding this rule
```
'no-undef': 'error',
```
to the eslint config file.

After adding axios this feature works!

PR #558, opened 2025-07-18
fix(pluralize): Fix pluralization issue in wizard

adonispuente:
Fixes a bug where the pluralization in the wizard reports incorrectly.
For testing, run the PR and go to Advisor, vulnerability, patch, or compliance to grab systems to remediate, and select the 'Plan remediation' button, there you should find the first instance of the pluralization fix (where the picture references). There are other instances of this but they require specific kinds of remediations. (manual types)

**With bug :** 
Screenshot 2025-07-18 at 12 43 30 PM

This PR fixes it so the number isnt duplicated as well as adds test to improve test coverage


bastilian:
@adonispuente I think we need to watch out and not commit the `build-tools` changes that are in here.

adonispuente:
@bastilian What are those? Ive noticed them populating a lot

src/modules/tests/RemediationsWizard.test.js
```diff
@@ -0,0 +1,231 @@
...
+  () => ({
+    __esModule: true,
+    useAxiosWithPlatformInterceptors: jest.fn(() => ({
+      get: () => {
+        let res = {
+          meta: {
+            count: 2,
+            total: 2,
+          },
+          data: [
```
bastilian, line 45: https://fakerjs.dev/guide/ ;)

adonispuente: hmmmm, im gonna set up another PR to get the app to like 50%, ill play around with faker in that. I completely forgot about it

bastilian: Was really just a hint that it could be used for generating some random data.

PR #567, opened 2025-07-25
Download

adonispuente:
Currently in remediations when downloading a single plan, it downloads a zip. This was a regression when updating to the JS Client. The Client currently doesnt support yml files but for the time being we should update the flow so that single downloads produce a yml file and bulk should produce a zip. 

You can download in the row items, bulk download in the main page, as well as the top right button in the details page


src/api/index.js
```diff
@@ -25,15 +25,24 @@ export function getHosts() {
...
+        ? downloadPlaybooksParamCreator({
+            selectedRemediations: selectedIds,
+            options: { baseURL: API_BASE },
+          }).then((res) => {
+            const { search, pathname } = res.urlObj;
+            const url = `${API_BASE}${pathname}${search}`;
+            window.location.assign(url);
+            return true;
+          })
+        : //TODO: Once JS Client is updated, replace with JS Client downloadPLaybooksParamCreator
```
bastilian, line 39: The JS clients are not updated automatically, so unless someone updates the remediations client no one will. ;)

bastilian: Does the OpenAPI definition still need updating? Otherwise it would just require running a task in the js clients repo and submit a PR with the updates?

src/api/index.js
```diff
@@ -23,26 +23,30 @@ export function getHosts() {
...
-        clearInterval(handle);
-        resolve();
-      }
-    }, 500);
-  });
-}
+export const downloadPlaybook = async (selectedIds) => {
+  try {
+    if (selectedIds.length > 1) {
+      downloadPlaybooksParamCreator({
```
adonispuente, line 29: @bastilian
The OpenAPI spec defines downloadPlaybooks as a GET, so the client hard-codes a GET withthe query params. this only “works” when you pass a single ID. As soon as you send multiple IDs, the generated query‐param format trips the endpoint (which in reality expects a POST with a JSON body).

src/api/index.js
```diff
@@ -23,26 +23,30 @@ export function getHosts() {
...
+        selectedRemediations: selectedIds,
+        options: { baseURL: API_BASE },
+      }).then((res) => {
+        const { search, pathname } = res.urlObj;
+        const url = `${API_BASE}${pathname}${search}`;
+        window.location.assign(url);
+        return true;
+      });
+    } else {
+      const res = await remediationsApi.downloadPlaybooks({
```
bastilian, line 39: Should this not be `remediationsApi.getRemediationPlaybook`?

bastilian: Also... you are calling the `remediationsApi.downloadPlaybooks` function as I think you describe is not working above? ... hm. fishier and fishier. :D

bastilian (approved):
Nice! Works as advertised! Thank you @adonispuente!

PR #599, opened 2025-08-14
feat(RHINENG-18943): Add status sorting/filtering

adonispuente:
This adds the ability to sort and filter by execution status on the main page (overview component). 
To test log in to remediations with the insights-qa account, sort columns by exec status, as well as try filtering by them. This also reenables the column manager


hugohezel (approved):
LGTM! Works nicely

adonispuente:
@hugohezel good catch, I added the failsafe for that so its not possible anymore

PR #609, opened 2025-08-20
feat(RHINENG-19641): Add rerouting and No permissions page

adonispuente:
This adds a 404 page for when an remediation ID results in a 404 on the details page (bookmarked a deleted plan)
Screenshot 2025-09-02 at 12 07 12 PM

Adds a no permissions page 
fixes the styling in PF6 No remediations page as well as updates the endpoint used to determine if the user has remediations or not. 

For testing reroute: you can use insights-qa 
Navigate to the details page of any remediation and then alter the id in the url to mimic a deleted plan, and see you are rerouted to the home page 

For testing **NoRemediations**: policies-noaccess-user-1 - with this user you can test a user with no remediations WITH permissions that can see the updated NoRemediationsPLans page styling (Note: This page is scoped by the orgs remediations, no if a user has remediations).
Screenshot 2025-08-26 at 11 13 52 AM

For the no access, if you cant alter/create a user, you can navigate to hasRequiredPermissions in App.js and locally change it to false
Screenshot 2025-08-26 at 11 14 33 AM

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -7,6 +7,7 @@ import { getIsReceptorConfigured } from './api';
 
 import NotificationsProvider from '@redhat-cloud-services/frontend-components-notifications/NotificationsProvider';
 import { Spinner } from '@patternfly/react-core';
+import { NotAuthorized } from '@redhat-cloud-services/frontend-components/NotAuthorized';
 
 export const PermissionContext = createContext();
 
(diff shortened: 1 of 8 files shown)


LightOfHeaven1994 (approved):
@adonispuente great work, looks good. 

Only 1 thing to consider (but it can be fixed in other PR) - When we navigate to details page and item is not found we usually show 404 page instead of rerouting (I want to believe it's a platform design. At least I see that Compliance and Inventory do this. So I would say we need to show ~"Remediation Plan is not found"
Screenshot 2025-08-28 at 10 56 28

adonispuente:
@LightOfHeaven1994 I like that better, im gonna implement that in this PR

adonispuente:
@LightOfHeaven1994 thats super odd, ill try to reproduce the issue. Right now im getting a 404 state

PR #611, opened 2025-08-27
feat(RHINENG=19595): Rebrand Insights to Red Hat Lightspeed

adonispuente:
For testing we're just looking for string changes. Navigate to every page and pop up you see and ctl+f for Insights to ensure nothing is there, as well as column names


LightOfHeaven1994:
@adonispuente Let's add feature flag 🥺

PR #614, opened 2025-09-02
fix(RHINENG-19750): Created Filter chip fix

adonispuente:
Theres an issue with the created filter chip where one you delete an input via backspace, the filterchip would still be present but with an empty value. This aims to resolve that as well as updates the test.

## Summary by Sourcery

Fix filter chip behavior to ignore empty or whitespace-only inputs, update selection logic accordingly, and ensure selected state defaults to an empty array while adding related unit tests.

Bug Fixes:
- Ignore empty or whitespace-only values in CalendarFilterType.chips to prevent empty filter chips
- Treat empty or whitespace-only input in CalendarFilterType.selectValue as deselection
- Default tableState.selected to an empty array in OverViewPage to avoid undefined

Tests:
- Add tests for CalendarFilterType.chips() and selectValue() handling of empty and whitespace-only strings

### DIFF
--- a/src/routes/OverViewPage/CalendarFilterType.js
+++ b/src/routes/OverViewPage/CalendarFilterType.js
@@ -3,11 +3,11 @@ import CalendarFilter from '../CalendarFilter';
 export const CalendarFilterType = {
   Component: CalendarFilter,
   chips: (value = []) => {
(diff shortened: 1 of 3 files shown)


hugohezel (approved):
LGTM, thank you!

PR #617, opened 2025-09-04
fix(RHINENG-20543): Provide nested table its own context

adonispuente:
[RHINENG-20543](https://issues.redhat.com/browse/RHINENG-20543)

In details page actions tab (on a plan with systems/actions in it), select the button that say 'x systems' on the row to open the modal. 

Opening that modal or filtering down on that modal should not affect the parent table at all.


src/routes/RemediationDetailsComponents/ActionsContent/SystemsModal/SystemsModal.js
```diff
@@ -68,7 +68,7 @@ SystemsModal.propTypes = {
 
 const SystemsModalContentProvider = (props) => {
   return (
-    <TableStateProvider>
+    <TableStateProvider isNewContext={true}>
```
bastilian, line 71: A boolean prop is automatically `true` if it is set even if it is not explicitly set with `={true}`. So just `<TableStateProvider isNewContext>` sh/would work.

bastilian: 🪿

PR #786, opened 2026-04-03
chore(packages): upgrade packages

adonispuente:
This just bulk updates outdated packages. Behavior in the app should not change in the slightest.

There should be absolutely no warnings in the terminal when build/running the application now.

## Summary by Sourcery

Upgrade dependencies and tooling and adjust config/tests to keep the app building cleanly on newer versions.

Enhancements:
- Configure module federation extensions in fec.config to satisfy the dynamic plugin webpack plugin and avoid runtime/build warnings.

Build:
- Upgrade application and devDependencies (React, PatternFly, Red Hat frontend components, tooling, and linters) to current compatible versions.

Tests:
- Tidy React test JSX usage and update Jest time-mocking calls to be compatible with the newer Jest version.


fec.config.js
hugohezel, line 56: We can update also these versions required for the host app.

hugohezel (commented):
Looks good to me, just one small thing

fec.config.js
adonispuente: Ive altered this to always match what the remediations app has, and made the required version be between 6 and 7

adonispuente: looking into this again, ive updated the version key to use what we have in package.json, however since we import this in 4 other apps, im still ok with having the range for the required version. I would only want to make that more strict if we upgraded react-router-dom here as a necessity and it broke another application if they didnt upgrade.

PR #867, opened 2026-07-07
feat(RHINENG-28527): add Amplitude analytics tracking for Remediation user actions

dajohnso:
In speaking with Talia, I threw this PR together as a prototype for the team to consider.  I think it looks fairly clean but maybe a little AI slop around proper importing of Chrome that I just wasn;t sure about nor the mocks for jest.  

Instrument 11 key user actions with chrome.analytics?.track() to enable understanding of plan creation, execution, deletion, downloads, and other high-value interactions.

# Description

Associated Jira ticket: https://redhat.atlassian.net/browse/RHINENG-28527

Please include a summary of the change, what this fixes/creates/improves.

# How to test the PR

Please include steps to test your PR.

# Before the change

# After the change

# Dependent work link

# Checklist:

## Summary by Sourcery

Instrument remediation workflows with analytics tracking for key user actions.

New Features:
- Track core remediation lifecycle events (creation, updates, execution, deletion, bulk deletion, rename) via chrome analytics.

Enhancements:
- Capture analytics for remediation-related UI interactions, including remediate button clicks, playbook downloads, action removals, and auto-reboot toggles.

Tests:
- Extend Jest chrome mocks to include analytics and auth to support the new tracking logic in tests.

Chores:
- Plumb chrome instance into remediation components and utilities that need to emit analytics events.

### DIFF
--- a/src/components/RenameModal.js
+++ b/src/components/RenameModal.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import TextInputDialog from './Dialogs/TextInputDialog';
 import { useAddNotification } from '@redhat-cloud-services/frontend-components-notifications/hooks';
+import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
 import useRemediations from '../Utilities/Hooks/api/useRemediations';
 import { updateRemediationWrapper } from '../routes/api';
 
@@ -15,6 +16,7 @@ const RenameModal = ({
   refetch,
 }) => {
   const addNotification = useAddNotification();
+  const chrome = useChrome();
   const { fetch: updateRemediation } = useRemediations(
     updateRemediationWrapper,
     {
@@ -27,6 +29,11 @@ const RenameModal = ({
       name = EMPTY_NAME;
     }
     const trimmedName = name.trim();
+
+    chrome.analytics?.track('remediations - Plan Renamed', {
+      module: 'remediations',
+      remediation_id: id,
+    });
 
     try {
       await updateRemediation({ id, name: trimmedName });
--- a/src/components/RenameModal.js
+++ b/src/components/RenameModal.js
@@ -30,13 +30,12 @@ const RenameModal = ({
     }
     const trimmedName = name.trim();
 
-    chrome.analytics?.track('remediations - Plan Renamed', {
-      module: 'remediations',
-      remediation_id: id,
-    });
-
     try {
       await updateRemediation({ id, name: trimmedName });
+      chrome.analytics?.track('remediations - Plan Renamed', {
+        module: 'remediations',
+        remediation_id: id,
+      });
       addNotification({
         title: `Remediation plan renamed`,
         variant: 'success',
--- a/src/modules/RemediationsButton.js
+++ b/src/modules/RemediationsButton.js
@@ -27,6 +27,7 @@ const RemediationButtonContent = ({
   isCompliancePrecedenceEnabled,
   buttonTooltipContent,
 }) => {
+  const chrome = useChrome();
   const [remediationsData, setRemediationsData] = useState();
   const [isNoDataModalOpen, setNoDataModalOpen] = useState(false);
 
@@ -65,6 +66,14 @@ const RemediationButtonContent = ({
 
                 try {
                   validate(data);
+                  chrome.analytics?.track(
+                    'remediations - Remediate Button Clicked',
+                    {
+                      module: 'remediations',
+                      issue_count: data?.issues?.length,
+                      system_count: data?.systems?.length,
+                    },
+                  );
                   setRemediationsData(data);
                 } catch {
                   setNoDataModalOpen(true);
--- a/src/modules/RemediationsButton.js
+++ b/src/modules/RemediationsButton.js
@@ -4,6 +4,7 @@ import propTypes from 'prop-types';
 import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
 import { AccessCheck } from '@project-kessel/react-kessel-access-check';
 import validate from './RemediationsModal/validate';
+import { normalizeRemediationData } from '../components/helpers';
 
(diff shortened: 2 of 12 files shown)


src/routes/RemediationDetailsComponents/ActionsContent/ActionsContent.js
```diff
@@ -216,6 +218,12 @@ const ActionsContent = ({
             }
             try {
               await handleDelete(chopped);
+              chrome.analytics?.track('remediations - Actions Removed', {
+                module: 'remediations',
+                remediation_id: id,
+                count: chopped.length,
+                is_bulk: isBulkDelete,
+              });
```
hugohezel, line 226: This tracker logs only after a successful deletion-action. In most of other places, the tracker logs before confirmation of a successful action (e.g. RemediationsWizard.js will log that plan has been created/updated even if the action failed, i.e. it will log an attempt rather than actual action). What is the tracking strategy in general that we should follow?

src/components/RenameModal.js
```diff
@@ -28,6 +30,11 @@ const RenameModal = ({
     }
     const trimmedName = name.trim();
 
+    chrome.analytics?.track('remediations - Plan Renamed', {
+      module: 'remediations',
+      remediation_id: id,
+    });
+
```
hugohezel: Remediations can be renamed also from Rem. Detail page > General tab > Details card. Therefore, this log should be added also here https://github.com/RedHatInsights/insights-remediations-frontend/blob/master/src/routes/RemediationDetailsComponents/DetailsCard.js#L137-L144 -> either into `try` block after `await`, or into `then` block before any `refetch` (if we want logging only successful actions. If we want to log all attempts, then its best to be placed into `try` block before `await`)

src/modules/RemediationsButton.js
```diff
@@ -65,6 +66,14 @@ const RemediationButtonContent = ({
...
                 try {
                   validate(data);
+                  chrome.analytics?.track(
+                    'remediations - Remediate Button Clicked',
+                    {
+                      module: 'remediations',
+                      issue_count: data?.issues?.length,
+                      system_count: data?.systems?.length,
+                    },
+                  );
```
hugohezel, line 78: Number of systems is likely incorrect here. This code only reads `data.systems.length`, but the remediation payload might be provided with systems nested under `data.issues[].systems`. The wizard already derives the displayed system count from both shapes using the `normalizeRemediationData()` helper.

Suggested changes:
1. import the helper:
`import { normalizeRemediationData } from '../components/helpers';`
2. use the helper:
```
...
try {
  validate(data);
  const normalizedData = normalizeRemediationData(data);
  chrome.analytics?.track('remediations - Remediate Button Clicked', {
    module: 'remediations',
    issue_count: normalizedData?.issues?.length ?? 0,
    system_count: normalizedData?.systems?.length ?? 0,
  });
  setRemediationsData(data);
} catch {
...
```

**Business note:** issues and systems counts represent the items passed into `RemediationsWizard` when it is opened. If the user later selects an existing plan in the wizard, these values may change because the wizard recalculates unique counts after merging the incoming items with the selected plan. So these counts may differ from the values reported when the user later clicks `Update plan` in the wizard.

hugohezel (commented):
Thank you @dajohnso. Looks good, just found a few things.

src/routes/RemediationDetailsComponents/ActionsContent/ActionsContent.js
dajohnso: Great call out Hugo.  I had claude investigate industry standards and best practice is to fire after API success by default for business-critical actions (CRUD operations like  create, delete, rename, execute). This tracks what actually happened, not just what was attempted.  It did say that sometimes you'd want to fire before to track intent but when getting started like we are, start with outcome only and add intent events when we have a specific question we are trying to ask, for example,  "why is our delete plan success rate so low".

src/components/RenameModal.js
dajohnso: addressed

src/modules/RemediationsButton.js
dajohnso: addressed

adonispuente (commented):
Here are 5 more places that was asked to be tracked on the Ticket that are currently not implemented in the PR:   

1. Download Preview Button (Create modal)
    - Function exists at RemediationWizard.js:309 (handlePreview)
    - No chrome.analytics?.track() call added
  2. Cancel Button (Create modal)
    - Button at RemediationWizard.js:454
    - Uses handleClose() which has no tracking
  3. Close (X) Button (Create modal)
    - Modal close handler at line 512
    - Also uses handleClose() - no tracking
  4. Launch Quick Start Button
    - Overview page: OverViewPageHeader.js:82
    - Calls quickStarts?.activateQuickstart() but no tracking
    - Note: Acceptance criteria mentions both "overview page" AND "execution readiness summary" locations
  5. AAP Trial Link ("Get a 30-day free trial...")
    - Link at DetailsGeneralContent.js:94-98
    - Plain <a> tag with no onClick handler or tracking
    - Shown when plan exceeds limits (>1000 actions or >100 systems)

dajohnso:
@adonispuente for the ProgressCard.js Quick Start tracking — the "Execution readiness summary" card has 3 separate activateQuickstart calls.  Did we want tracking from these or maybe its overkill for now and we can re-vist after we start getting some data.  Maybe that is more a question for Suresh and/or Talia?

  1. RBAC/permissions popover — "Open the quick start" button (~line 144)
  2. Connected systems popover — "Open the quick start" button (~line 200)
  3. "Learn more" link in the card body (~line 248)

adonispuente:
@dajohnso Im alright with sending things up as is now, and if Talia or anyone wants to add more we can always do so then. It might be overkill to track all 3

adonispuente (approved):
/lgtm, thank you!

